PR #14, opened 2024-05-29
Add LiteLLM router to FastAPI container

petermuller:
This set of changes adds a dependency on LiteLLM in our FastAPI container so that we may pass requests from the LISA Serve ALB directly to LiteLLM. This enables us to handle any form of OpenAI API spec implementation so long as LiteLLM also supports it.

Summary of changes:
1. Add LiteLLM dependency
2. Add /v2/serve route to the FastAPI container for the new implementation
3. Update the schema file to embed the entire LiteLLM config into the LISA config yaml file
4. Add LiteLLM config snippet to the example_config.yaml
5. Add runtime logic for adding LISA-served models to the LiteLLM on container start

Testing:
1. Deployed a personal stack using the TGI and TEI container models from the example config
2. Validated that I could perform OpenAI spec operations on my backend ALB and received intelligible responses from the requests
```
curl -H 'Api-Key: myRedactedKey' -X GET https://myredacteddomain/v2/serve/models
curl -H 'Api-Key: myRedactedKey' -X POST https://myredacteddomain/v2/serve/chat/completions -d '{"model":"mistralai/Mistral-7B-Instruct-v0.2","messages":[{"role":"user","content":"You are a helpful AI assistant who responds to user requests in a concise and accurate manner. All of the following is a conversation between you and the user, and you respond helpfully and thoughtfully to the user. Your responses should be as short as possible."},{"role":"assistant","content":"Understood."},{"role":"user","content":"Suggest a programming language and a project to complete in it."}], "stream": true, "max_tokens": 1500}'
```


petermuller:
I'll add a new commit to fix the failing tests.

lib/schema.ts
```diff
@@ -663,6 +663,81 @@ const ApiGatewayConfigSchema = z
...
+      output_cost_per_token: z.number().optional(),
+      max_tokens: z.number().optional(),
+      base_model: z.string().optional(),
+    })
+    .optional(),
+});
+
+/**
+ * Core LiteLLM configuration.
+ * See https://litellm.vercel.app/docs/proxy/configs#all-settings for more details about each field.
```
KyleLilly, line 692: I think this is fine in general but when we add some examples to the README we may want to add some additional details about the properties we expect customer may want to actually set. You've got comments on some parameters but not most and some of them seem specific to google/azure which wouldn't apply at all for LISA.

lib/schema.ts
```diff
@@ -776,6 +851,7 @@ const RawConfigSchema = z
       })
       .optional(),
     stackSynthesizer: z.nativeEnum(stackSynthesizerType).optional(),
+    litellmConfig: LiteLLMConfig,
```
KyleLilly, line 854: This may not apply until you've fully migrated things but at the moment `ecsModels` is still required (it will also fail on `make deploy` due to the `modelCheck` target), if I want to just use SageMaker hosted endpoints I don't have a way to do that.

lib/serve/rest-api/src/api/endpoints/v2/litellm_passthrough.py
```diff
@@ -0,0 +1,64 @@
...
+import logging
+from collections.abc import Iterator
+from typing import Union
+
+import requests
+from fastapi import APIRouter, Request
+from fastapi.responses import JSONResponse, Response, StreamingResponse
+
+# Local LiteLLM installation URL
+LITELLM_URL = "http://localhost:4000"
```
KyleLilly, line 26: Does this need to be more dynamic? If not should we add a note in the entrypoint script that if the port is changed for litellm this also needs to be updated?

lib/serve/rest-api/src/api/routes.py
```diff
@@ -30,6 +31,9 @@
 router.include_router(models.router, prefix="/v1", tags=["models"], dependencies=[Depends(security)])
 router.include_router(embeddings.router, prefix="/v1", tags=["embeddings"], dependencies=[Depends(security)])
 router.include_router(generation.router, prefix="/v1", tags=["generation"], dependencies=[Depends(security)])
+router.include_router(
+    litellm_passthrough.router, prefix="/v2/serve", tags=["litellm_passthrough"], dependencies=[Depends(security)]
```
KyleLilly, line 35: I'm assuming this is `/v2/serve` because we're going to move the API GW stuff into `/v2/` as well?

lib/serve/rest-api/src/requirements.txt
```diff
@@ -1,10 +1,13 @@
 aiobotocore
```
KyleLilly, line 1: Not directly related to this PR but we need to pin these.

lib/serve/rest-api/src/entrypoint.sh
```diff
@@ -5,6 +5,14 @@ set -e
 HOST="0.0.0.0"
 PORT="8080"
 
+# Update LiteLLM config that was already copied from config.yaml with runtime-deployed models.
+# Depends on SSM Parameter for registered models.
+echo "Configuring and starting LiteLLM"
+python ./src/utils/generate_litellm_config.py -f litellm_config.yaml
```
KyleLilly, line 11: Similar to my other comment we should put a comment here about how this file is generated from the `Dockerfile` and the name shouldn't change here unless it changes there.

lib/schema.ts
petermuller: ack. I didn't want to restrict folks' abilities to use the entire LiteLLM config- I only have as much information as the proxy docs have (which is about as accurate as I can be). I can add docs to the readme for this.

lib/schema.ts
petermuller: Good call, I will send a new revision with that.

lib/serve/rest-api/src/api/endpoints/v2/litellm_passthrough.py
petermuller: I can write that comment- this is the only thing running alongside the FastAPI, and since the port isn't exposed, I'm not expecting anything to call it directly or have a need for it to be changed.

lib/serve/rest-api/src/api/routes.py
petermuller: Yes, that is correct. I'm also welcome to changing the name, but I wanted to avoid the situation of calling it "model" and having a url like the following when using it directly for the OpenAI list models call:

```
curl -X GET https://mydomain/v2/model/models
```

lib/serve/rest-api/src/requirements.txt
petermuller: Not a problem, I'll generate those

lib/serve/rest-api/src/entrypoint.sh
petermuller: Will do, thanks!

lib/schema.ts
petermuller: Just tried this- this is going to be a much larger effort, I'm going to create an issue for it to follow up on it later.

lib/serve/rest-api/src/api/routes.py
```diff
@@ -30,6 +31,9 @@
 router.include_router(models.router, prefix="/v1", tags=["models"], dependencies=[Depends(security)])
```
estohlmann, line 31: Are we going to be marking the v1 endpoints as depreciated?

estohlmann (approved):

lib/serve/rest-api/src/api/routes.py
petermuller: There's not going to be a lot of time between them being deprecated and us releasing the next revision, so I think for now, we can just leave as-is

PR #16, opened 2024-06-04
Add support for vLLM containers

petermuller:
This commit adds support for vLLM containers with some example documentation to support it. The schema has been updated to allow for the vLLM container. These changes help toward the migration to the v2.x architecture.

*Description of changes:*
1. Update README with examples of usage
2. Update schema to guarantee no periods in names so that I can split on them later
3. Add Dockerfile and entrypoint script for vLLM-served models
4. Add a short-circuit for non-TGI/TEI containers so that we don't fail on "registering" the model, which will be deprecated as part of an upcoming v2 release.

### DIFF
--- a/lib/serve/ecs-model/vllm/src/entrypoint.sh
+++ b/lib/serve/ecs-model/vllm/src/entrypoint.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+set -e
+
+declare -a vars=("S3_BUCKET_MODELS" "LOCAL_MODEL_PATH" "MODEL_NAME" "S3_MOUNT_POINT" "THREADS")
+
+# Check the necessary environment variables
+for var in "${vars[@]}"; do
+    if [[ -z "${!var}" ]]; then
+        echo "$var must be set"
+        exit 1
+    fi
+done
+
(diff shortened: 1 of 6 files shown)


lib/serve/ecs-model/vllm/src/entrypoint.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+set -e
+
+declare -a vars=("S3_BUCKET_MODELS" "LOCAL_MODEL_PATH" "MODEL_NAME" "S3_MOUNT_POINT" "THREADS")
+
+# Check the necessary environment variables
```
estohlmann, line 6: Are these variables always apart of these base images? or is this something that we should make sure we add to the README?

petermuller: These are provided from the container build params within the typescript definition, so they should all be defined as part of this block: https://github.com/awslabs/LISA/blob/646e0fa057a5c6aa76acb7acf151b4b58b649ae1/lib/serve/ecs-model/index.ts#L86-L123

estohlmann (approved):

PR #17, opened 2024-06-04
Update FE to comply with OpenAI Spec

estohlmann:
Updating the FE to use new V2 Open AI endpoints 

*Description of changes:*
1. Switch over to using @langchain/OpenAI Chat library instead of the custom LISA adapter that was tailored to TGI
2. Updated to widely supported [Chat API](https://platform.openai.com/docs/api-reference/chat/create) vs legacy [Completions API](https://platform.openai.com/docs/api-reference/completions/create)
3. Updated ModelKwargs to support nullable OpenAI fields and their API field descriptions
4. Removed un-needed custom object types the UI was tightly coupled to ie `ModelProvider`
5. Update configuration documentation references to point to v2 endpoints


lib/user-interface/react/src/components/chatbot/Chat.tsx
```diff
@@ -293,6 +281,24 @@ export default function Chat({ sessionId }) {
     };
   };
 
+  const createOpenAiClient = (streaming: boolean) => {
+    return new ChatOpenAI({
+      modelName: selectedModel?.id,
+      openAIApiKey: auth.user?.id_token,
+      configuration: {
+        baseURL: `${RESTAPI_URI}/v2/serve`,
```
KyleLilly, line 289: This should probably be using the `RESTAPI_VERSION` variable?

lib/user-interface/react/src/components/chatbot/Chat.tsx
```diff
@@ -319,41 +325,10 @@ export default function Chat({ sessionId }) {
...
-      idToken: auth.user?.id_token,
-      modelName: selectedModel.modelName,
-      providerName: selectedModel.provider,
-      modelKwargs: modelKwargs,
-      streaming: false,
-      contentHandler: new LisaContentHandler(),
-    });
+
+    const llm = createOpenAiClient(streamingEnabled);
+    const llmNoCallback = createOpenAiClient(false);
```
KyleLilly, line 330: Now that neither of them are using a callback and the logic for updating the chat session data is part of the chain it would make it more clear to rename this to indicate it's used for prompt contextualization. It may be possible to remove it all together too.

lib/user-interface/react/src/components/utils.ts
```diff
@@ -170,20 +166,11 @@ export const deleteUserSessions = async (idToken: string) => {
...
-  idToken: string,
-): Promise<DescribeModelsResponseBody> => {
-  const queryParams: string = modelTypes.map((elem) => `modelTypes=${elem}`).join('&');
-  const resp = await sendAuthenticatedRequest(
-    `${RESTAPI_URI}/${RESTAPI_VERSION}/describeModels?${queryParams}`,
-    'GET',
-    idToken,
-  );
+export const describeModels = async (idToken: string): Promise<DescribeModelsResponseBody> => {
+  const resp = await sendAuthenticatedRequest(`${RESTAPI_URI}/v2/serve/models`, 'GET', idToken);
```
KyleLilly, line 173: Similar comment about removing the use of `RESTAPI_VERSION`. If it's going to be removed and just hardcoded you could update the `example_config.yaml` and the CDK code where the `env.js` is created to remove the parameter there too.

lib/user-interface/react/src/components/chatbot/Chat.tsx
estohlmann: I had these set to `RESTAPI_VERSION` but @petermuller had some concern about customers having to know to update their configuration files when doing the redeploy and that potentially causing issues.

lib/user-interface/react/src/components/chatbot/Chat.tsx
estohlmann: Good call, I can embed this directly into 'contextualizeQChain'

lib/user-interface/react/src/components/chatbot/Chat.tsx
estohlmann: Talked with peter, I think we are good with keeping `RESTAPI_VERSION` - adding it back to the URLs

lib/rag/layer/requirements.txt
```diff
@@ -2,6 +2,7 @@ opensearch-py
 requests-aws4auth
 PyPDF2==3.0.1
 langchain==0.1.0
+langchain-openai==0.0.6
```
petermuller, line 5: Just a heads up that we may want to look into updating these since they're a bit old and the OpenAIEmbeddings client is on a deprecation path. Not a blocker for now (especially since I'm the one that committed that), but will possibly do a pass before official release to ensure we're not too old on requirements versions.

test/cdk/mocks/config.yaml
```diff
@@ -54,7 +54,7 @@ dev:
   apiGatewayConfig:
     domainName:
   restApiConfig:
-    apiVersion: v1
+    apiVersion: v2
```
petermuller, line 57: Do we need to change this in the example_config.yaml too?

petermuller: nvm you already did. disregard this.

lambda/repository/lambda_functions.py
```diff
@@ -68,18 +68,18 @@ def _get_cert_path() -> str | bool:
...
         lisa_api_endpoint = lisa_api_param_response["Parameter"]["Value"]
 
-    lisa = Lisa(
-        url=lisa_api_endpoint, verify=_get_cert_path(), timeout=60, headers={"Authorization": f"Bearer {id_token}"}
-    )
+    headers = {"Authorization": f"Bearer {id_token}"}
 
-    embedding = LisaEmbeddings(provider=provider, model_name=model_name, client=lisa)
+    embedding = LisaOpenAIEmbeddings(
+        lisa_openai_api_base=lisa_api_endpoint + "/v2/serve", model=model_name, headers=headers
```
petermuller, line 81: should we get the apiVersion here from the environment variables (do we have a way to get that, either in the env or in an SSM Param)?

lib/user-interface/react/src/components/chatbot/ModelKwargs.tsx
```diff
@@ -29,62 +27,43 @@ import {
...
-  const [topP, setTopP] = useState(0.99); // TGI client enforces strictly less than 1.0
-  const [typicalP, setTypicalP] = useState(0.99); // TGI client enforces strictly less than 1.0
-  const [temperature, setTemperature] = useState(1.0);
-  const [repetitionPenalty, setRepetitionPenalty] = useState(1.0);
-  const [returnFullText, setReturnFullText] = useState(false);
-  const [watermark, setWatermark] = useState(false);
-  const [doSample, setDoSample] = useState(false);
-  const [truncate, setTruncate] = useState(1024);
-  const [seed, setSeed] = useState(42);
-  const [stopSequences, setStopSequences] = useState(['\nUser:', '\n User:', 'User:', 'User']);
```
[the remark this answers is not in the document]
estohlmann: I can easily add these back in

lambda/repository/lambda_functions.py
petermuller: I pushed a change to add an environment variable for this. Resolving thread.

petermuller (approved):

PR #18, opened 2024-06-04
APIGW -> ALB Migration

douglas1850:
*Description of changes:*

* Move the ALB created in the ecsCluster stack into a common place where it can be re-used. A public LB is created for FastAPI and other public API endpoints, with a new private ALB being created to specifically handle private model endpoints.

TODO:
* create endpoints for the session and RAG endpoints
* create an endpoint for website hosting


petermuller:
There are a lot of things in flux right now, so it's going to be easier if we wrap this into another feature implementation or as a set of changes in a newer feature branch. I'm going to close this as we're not actively looking into this at the moment, but we still may revisit later.

PR #19, opened 2024-06-05
Split textgen and embedding models in UI

estohlmann:
*Description of changes:*
- Updated UI to pull Models from config vs API, this allows us to get supplemental info
- UI now only separates textgen vs embedding models in dropdowns, also only enables streaming for models that support it
- Fixed longstanding bug where switching to a model with streaming active to a model that doesn't support streaming "locks" streaming as enabled

demo

### DIFF
--- a/README.md
+++ b/README.md
@@ -415,7 +415,24 @@ window.env = {
   // Alternatively you can set this to be your REST api elb endpoint
   RESTAPI_URI: 'http://localhost:8080/',
   RESTAPI_VERSION: 'v2',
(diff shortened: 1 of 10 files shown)


estohlmann:
update to call the describe API and uses the results to map against the config. I was a little forward looking with my approach and this accounts for the following:
- If a model is returned in the API that isn't in the config it will add it to both drop-downs and allow streaming to be toggled on
- If a model isn't returned that IS listed in the config, it will NOT show up in the dropdowns

petermuller (approved):

PR #20, opened 2024-06-06
Documentation Updates for v2

petermuller:
Update README for the updates that we are pushing for LISA V2.

Includes:
* New architecture diagrams
* vLLM docs
* OpenAI Compatibility usage and examples
* LiteLLM configuration + SageMaker and Bedrock documentation items were added in a previous commit: lines 93-112 if you would like to expand and provide feedback on those too


README.md
```diff
@@ -443,6 +454,94 @@ cd lib/user-interface/react/
...
+curl -s -H 'Api-Key: your-api-token' -X GET https://${lisa_serve_alb}/v2/serve/models
+```
+
+If using the IdP, the request would look like the following:
+
+```shell
+curl -s -H 'Authorization: Bearer your-bearer-token' -X GET https://${lisa_serve_alb}/v2/serve/models
+```
+
+When using a library that requests an OpenAI-compatible base_url, you can provide `https://${lisa_serve_alb}/v2/serve` here. All of the OpenAI routes will
```
estohlmann, line 487: Just want to make sure `lisa_serve_alb` is still corect with the ALB complications we had.

estohlmann (approved):

README.md
petermuller: Yes, these requests still go to the load balancer and not the APIGW endpoint

PR #37, opened 2024-06-20
Set default penalty params to null when unspecified

petermuller:
The LangChain JS library sets some default values for presence and frequency penalties if we don't set them as part of the OpenAI client. As a result, we end up with a non-required numeric value (0) that we did not intend to put there, which heavily degrades the model responses on tgi containers. These changes override the defaults set by the client so that the values are either the numeric value we set, including 0, otherwise they will be null. I validated that this happens by setting and unsetting the values in the model kwargs of the Chat UI and checked the request data from the network tab of my browser. Now, by default, these values won't be set unless we want them to be.

### DIFF
--- a/lib/user-interface/react/src/components/chatbot/Chat.tsx
+++ b/lib/user-interface/react/src/components/chatbot/Chat.tsx
@@ -279,7 +279,7 @@ export default function Chat({ sessionId }) {
   };
 
   const createOpenAiClient = (streaming: boolean) => {
-    const client = new ChatOpenAI({
+    return new ChatOpenAI({
       modelName: selectedModel?.id,
       openAIApiKey: auth.user?.id_token,
       configuration: {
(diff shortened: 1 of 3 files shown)


lib/user-interface/react/src/components/chatbot/Chat.tsx
```diff
@@ -295,6 +295,16 @@ export default function Chat({ sessionId }) {
       stop: modelConfig?.stop,
       modelKwargs: modelConfig?.modelKwargs,
     });
+    // LangChain's JS OpenAI client sets default values that we do not want if we didn't set them.
+    // The following parameters are preferred to be null instead of having a numeric default value
+    // as they degrade the quality of the model response if set incorrectly by default.
+    if (!modelConfig?.frequency_penalty && modelConfig?.frequency_penalty !== 0) {
```
estohlmann, line 301: We should be able to add these inline above like this:

`
new ChatOpenAI({
      modelName: selectedModel?.id,
      openAIApiKey: auth.user?.id_token,
      configuration: {
        baseURL: `${RESTAPI_URI}/${RESTAPI_VERSION}/serve`,
      },
      streaming,
      maxTokens: modelConfig?.max_tokens,
      n: modelConfig?.n,
      topP: modelConfig?.top_p,
      frequencyPenalty: modelConfig?.frequency_penalty ?? null,
      presencePenalty: modelConfig?.presence_penalty ?? null,
      temperature: modelConfig?.temperature,
      stop: modelConfig?.stop,
      modelKwargs: modelConfig?.modelKwargs,
    });
`

estohlmann, line 301: I sent you a patch file but what I would suggest doing is pulling the properties that we want to null out by defualt out into the modelKwargs property of the modelConfig. These properties will be defaulted to null when passed into modelKwargs and you will get the behavior you are replicating here in a much cleaner way.

petermuller: Awesome suggestion- doing that for the next set of changes. I left the number of tokens alone because those are usually required at the client level, but this is a much nicer approach and we don't need any other custom logic.

estohlmann (approved):

PR #42, opened 2024-07-30
Feat/private lisa serve

krzim-aws:
This PR adds the ability to selectively deploy the various components of LISA, allows for multiple deployments of LISA into a single account, and allows for deployment of LISA-Serve without auth inside a VPC.

### DIFF
--- a/README.md
+++ b/README.md
@@ -359,7 +359,7 @@ An account owner may create a long-lived API Token using the following AWS CLI c
 ```bash
 AWS_REGION="us-east-1"  # change to your deployment region
 token_string="YOUR_STRING_HERE"  # change to a unique string for a user
-aws --region $AWS_REGION dynamodb put-item --table-name LISAApiTokenTable \
+aws --region $AWS_REGION dynamodb put-item --table-name <deployment name>-LISAApiTokenTable \
     --item '{"token": {"S": "'${token_string}'"}}'
 ```
 
@@ -371,7 +371,7 @@ in seconds. The following command shows an example of how to do this.
 AWS_REGION="us-east-1"  # change to your deployment region
 token_string="YOUR_STRING_HERE"
 token_expiration=$(echo $(date +%s) + 3600 | bc)  # token that expires in one hour, 3600 seconds
-aws --region $AWS_REGION dynamodb put-item --table-name LISAApiTokenTable \
+aws --region $AWS_REGION dynamodb put-item --table-name <deployment name>-LISAApiTokenTable \
     --item '{
         "token": {"S": "'${token_string}'"},
         "tokenExpiration": {"N": "'${token_expiration}'"}
@@ -401,7 +401,7 @@ that key.
 AWS_REGION="us-east-1"  # change to your deployment region
 token_string="YOUR_STRING_HERE"
 token_expiration=$(echo $(date +%s) + 600 | bc)  # token that expires in 10 minutes from now
-aws --region $AWS_REGION dynamodb update-item --table-name LISAApiTokenTable \
+aws --region $AWS_REGION dynamodb update-item --table-name <deployment name>-LISAApiTokenTable \
     --key '{"token": {"S": "'${token_string}'"}}' \
     --update-expression 'SET tokenExpiration=:t' \
     --expression-attribute-values '{":t": {"N": "'${token_expiration}'"}}'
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -359,7 +359,7 @@ An account owner may create a long-lived API Token using the following AWS CLI c
 ```bash
 AWS_REGION="us-east-1"  # change to your deployment region
 token_string="YOUR_STRING_HERE"  # change to a unique string for a user
-aws --region $AWS_REGION dynamodb put-item --table-name LISAApiTokenTable \
+aws --region $AWS_REGION dynamodb put-item --table-name <deployment name>-LISAApiTokenTable \
```
petermuller, line 362: To nitpick on these, can you make the deployment be a variable so it's copy/paste-able (and same with the rest of the commands). it's ok if the variable is hardcoded to be `deployment=dev`

lib/networking/vpc/index.ts
```diff
@@ -118,8 +118,13 @@ export class Vpc extends Construct {
     // All HTTP VPC traffic -> ECS model ALB
     ecsModelAlbSg.addIngressRule(Peer.ipv4(vpc.vpcCidrBlock), Port.tcp(80), 'Allow VPC traffic on port 80');
 
-    // All HTTPS IPV4 traffic -> REST API ALB
-    restApiAlbSg.addIngressRule(Peer.anyIpv4(), Port.tcp(443), 'Allow any traffic on port 443');
+    if (config.restApiConfig.internetFacing) {
```
petermuller, line 121: should this be based on the existence of the cert or not on the rest api alb? i can still have a cert on a private/internal ALB, which i'd probably want to listen on 443.

README.md
krzim-aws: Can do!

lib/schema.ts
```diff
@@ -840,21 +846,23 @@ const RawConfigSchema = z
       })
       .optional(),
     deployRag: z.boolean().optional().default(false),
+    deployChat: z.boolean().optional().default(false),
+    deployUi: z.boolean().optional().default(false),
```
petermuller, line 850: This would potentially be a breaking change for people who have already deployed LISA with default options. For the next release, we can add notes to ensure people specify these if they've previously deployed, just so they have a breaking change warning.

README.md
krzim-aws: Ah actually the envvar `DEPLOYMENT_NAME` is defined as a part of the setup previously in the README. I'll update these to use that.

lib/networking/vpc/index.ts
krzim-aws: Good call. I'll make that adjustment.

lib/schema.ts
krzim-aws: I can make them default to `true` which should eliminate the breaking change issue, right?

krzim-aws: Admittedly, this was likely just a copy/paste oversight from the `deployRag` schema.

lib/schema.ts
```diff
@@ -826,6 +831,7 @@ const RawConfigSchema = z
         message: 'AWS account number should be 12 digits',
       }),
     region: z.string(),
+    stack: z.string().default('all'),
```
petermuller, line 834: Is this used?

krzim-aws: No. Deleted.

lib/schema.ts
```diff
@@ -840,21 +846,23 @@ const RawConfigSchema = z
       })
       .optional(),
     deployRag: z.boolean().optional().default(false),
+    deployChat: z.boolean().optional().default(false),
+    deployUi: z.boolean().optional().default(false),
     logLevel: z.union([z.literal('DEBUG'), z.literal('INFO'), z.literal('WARNING'), z.literal('ERROR')]),
     lambdaConfig: lambdaConfigSchema,
     lambdaSourcePath: z.string().optional().default('./lambda'),
-    authConfig: AuthConfigSchema,
+    authConfig: AuthConfigSchema.optional(),
```
petermuller, line 853: can we also add a validation for the authconfig being defined if we do end up using the chat stacks? The UI won't work otherwise in its current state (it'll deploy, but we won't be able to use it)

lib/schema.ts
```diff
@@ -794,6 +796,9 @@ const LiteLLMConfig = z.object({
  * @property {string} s3BucketModels - S3 bucket for models.
  * @property {string} mountS3DebUrl - URL for S3-mounted Debian package.
  * @property {string[]} [accountNumbersEcr=null] - List of AWS account numbers for ECR repositories.
+ * @property {boolean} [deployRag=false] - Whether to deploy RAG stacks.
+ * @property {boolean} [deployChat=false] - Whether to deploy chat stacks.
+ * @property {boolean} [deployUi=false] - Whether to deploy UI stacks.
```
petermuller, line 801: minor nitpick on these since they're defaulting to `true` now. thanks for adjusting those- that will help reduce a lot of confusion from someone accidentally deleting the UI :)

petermuller (approved):
Thanks for the changes!

PR #45, opened 2024-08-13
Add PostgreSQL db for LiteLLM scaling

petermuller:
This set of changes adds a PostgreSQL database to the Serve infrastructure to help with scaling LiteLLM across multiple instances. This database will be used between multiple REST API containers so that models that are dynamically added to a running LiteLLM instance will update the DB and be available across all other REST API containers behind the LISA Serve endpoint.

Tests have increased error counts because the database is a Single-AZ db. This is expected to be fine because the database is only used for this one purpose and is not anticipated to have any form of high load on it.

This database will be used in further model management feature development.

### DIFF
--- a/lib/serve/rest-api/src/utils/generate_litellm_config.py
+++ b/lib/serve/rest-api/src/utils/generate_litellm_config.py
@@ -22,6 +22,7 @@ import click
 import yaml
 
 ssm_client = boto3.client("ssm", region_name=os.environ["AWS_REGION"])
+secrets_client = boto3.client("secretsmanager", region_name=os.environ["AWS_REGION"])
 
 
 @click.command()  # type: ignore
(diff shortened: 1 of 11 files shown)


lib/serve/rest-api/src/utils/generate_litellm_config.py
```diff
@@ -51,6 +52,22 @@ def generate_config(filepath: str) -> None:
     config_models = config_contents["model_list"] or []  # ensure config_models is a list and not None
     config_models.extend(litellm_model_params)
     config_contents["model_list"] = config_models
+
+    # Get database connection info
+    db_param_response = ssm_client.get_parameter(Name=os.environ["LITELLM_DB_INFO_PS_NAME"])
+    db_params = json.loads(db_param_response["Parameter"]["Value"])
+    secrets_response = secrets_client.get_secret_value(SecretId=db_params["passwordSecretId"])
+    password = json.loads(secrets_response["SecretString"])["password"]
+    connection_str = f"postgresql://{db_params['username']}:{password}@{db_params['dbHost']}:5432/{db_params['dbName']}"
```
dustins, line 61: I know this is pretty standard but should we make the port configurable too?

petermuller: Yea I can do that- I'll update the schema to add it there by default- since I'm using the same rds config as the RAG repo- I'll need to update there too.

lib/serve/rest-api/src/utils/generate_litellm_config.py
```diff
@@ -22,6 +22,7 @@
 import yaml
 
 ssm_client = boto3.client("ssm", region_name=os.environ["AWS_REGION"])
+secrets_client = boto3.client("secretsmanager", region_name=os.environ["AWS_REGION"])
```
estohlmann, line 25: Do we need to update any architecture call outs that we are taking on a new dependency for secretsmanager... Or did we already use this before?

petermuller: We already use secrets manager in the RAG operations where we need to use the postgres db there too. We've used it before 👍

estohlmann (approved):

PR #54, opened 2024-08-27
Create and Update Model Form

estohlmann:
*Issue #, if available:*

*Description of changes:*

Adding the ability for users to create/update models via the LISA app if they are admins. A quick demo of the form can be seen below. *Note*: there are still placeholder descriptions in some places.

output


lib/user-interface/react/src/shared/model/model-management.model.ts
```diff
@@ -29,11 +31,142 @@ export enum ModelType {
...
+};
+
+export type IContainerConfig = {
+    BaseImage: IContainerConfigImage;
+    SharedMemorySize: number;
+    HealthCheckConfig: IContainerHealthCheckConfig;
+    Environment?: Record<string, string>[];
+};
+
 export type IModel = {
```
petermuller, line 89: I'll have to edit this in my PR, so don't worry about for now, but I'm also adding a "UniqueId" which is the UUID that comes from LiteLLM, which is how we'd eventually reference the model for modifications/mutations if we need to do anything from the LiteLLM perspective.

lib/user-interface/react/src/shared/model/model-management.model.ts
```diff
@@ -29,11 +31,142 @@ export enum ModelType {
...
     ModelType: ModelType;
-    ModelUrl: string;
+    InstanceType: string;
+    InferenceContainer: InferenceContainer;
+    ContainerConfig: IContainerConfig;
+    AutoScalingConfig: IAutoScalingConfig;
+    LoadBalancerConfig: ILoadBalancerConfig;
+};
+
+export type IModelRequest = {
```
petermuller, line 101: No action to take, but also to note (primarily for myself), this will need to have the ModelUrl added as a potential field as well. In the case that we add something directly to LiteLLM and don't create the infrastructure for it, we'd need the ability to set the model URI too.

petermuller (approved):

PR #56, opened 2024-08-28
Move ECS model infrastructure to separate stack, deploy from Lambda

stephensmith-aws:
Move ECS model deployment into nested lambda. These stacks get deployed by invoking the lambda.
Example event:
```
{
  "action": "destroy",
  "modelConfig": {
    "modelName": "mistralai/Mistral-7B-Instruct-v0.2",
    "modelId": "mistral7b",
    "deploy": true,
    "streaming": true,
    "modelType": "textgen",
    "instanceType": "g5.xlarge",
    "inferenceContainer": "tgi",
    "containerConfig": {
      "image": {
        "baseImage": "ghcr.io/huggingface/text-generation-inference:2.0.1",
        "path": "lib/serve/ecs-model/textgen/tgi",
        "type": "asset"
      },
      "sharedMemorySize": 2048,
      "healthCheckConfig": {
        "command": [
          "CMD-SHELL",
          "exit 0"
        ],
        "interval": 10,
        "startPeriod": 30,
        "timeout": 5,
        "retries": 3
      },
      "environment": {
        "MAX_CONCURRENT_REQUESTS": 128,
        "MAX_INPUT_LENGTH": 1024,
        "MAX_TOTAL_TOKENS": 2048
      }
    },
    "autoScalingConfig": {
      "minCapacity": 1,
      "maxCapacity": 1,
      "defaultInstanceWarmup": 180,
      "cooldown": 420,
      "metricConfig": {
        "AlbMetricName": "RequestCountPerTarget",
        "targetValue": 30,
        "duration": 60,
        "estimatedInstanceWarmup": 330
      }
    },
    "loadBalancerConfig": {
      "healthCheckConfig": {
        "path": "/health",
        "interval": 60,
        "timeout": 30,
        "healthyThresholdCount": 2,
        "unhealthyThresholdCount": 10
      }
    }
  }
}
```


ecs_model_deployer/.gitignore
```diff
@@ -0,0 +1,2 @@
+dist/
```
petermuller, line 1: Is a new .gitignore needed for the project? Is the toplevel one not sufficient?

lib/serve/index.ts
```diff
@@ -134,42 +133,10 @@ export class LisaServeApplicationStack extends Stack {
...
-                // For textgen models, add metadata whether streaming is supported
-                if (modelConfig.modelType === ModelType.TEXTGEN) {
-                    registeredModel.streaming = modelConfig.streaming!;
-                }
-                registeredModels.push(registeredModel);
-            }
-        }
-
         // Create Parameter Store entry with registeredModels
         this.modelsPs = new StringParameter(this, createCdkId(['RegisteredModels', 'StringParameter']), {
```
petermuller, line 137: We can remove this if it's not going to be used anymore, unless we need it in other areas.

lib/models/ecs-model-deployer.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export class ECSModelDeployer extends Construct {
+    constructor (scope: Construct, id: string, props: ECSModelDeployerProps) {
+        super(scope, id);
+        const stackName = Stack.of(scope).stackName;
+        const role = new Role(this, createCdkId([stackName, 'ecs-model-deployer-role']), {
+            assumedBy: new ServicePrincipal('lambda.amazonaws.com')
+        });
+
+        role.addManagedPolicy(ManagedPolicy.fromAwsManagedPolicyName('AdministratorAccess'));
```
petermuller, line 39: this is probably going to get flagged for being overly permissive. since its cloudformation and we need a ton of permissions anyways, could we at least restrict to the "*" policies for specific services we use? like `ec2:*`, `ecs:*`, `lambda:*`, and so on?

ecs_model_deployer/src/index.ts
```diff
@@ -0,0 +1,130 @@
...
+        }
+        const source = `/tmp/${f}`;
+        const target = `${process.env['PWD']}/${f}`;
+
+        try {
+            symlinkSync(target, source, 'file');
+        } catch ( err ) {
+            if ( err instanceof Error && err.message.match(/EEXIST/) ) {
+                // Writable env already established from previous call.
+            }
```
petermuller, line 47: the outcome is the same whether we hit the condition or not- should we throw another error, or just log that we're skipping if there's an exception?

lib/models/ecs-model-deployer.ts
```diff
@@ -0,0 +1,68 @@
...
+            'mountS3DebUrl': props.config.mountS3DebUrl
+        };
+
+        const functionId = createCdkId([stackName, 'ecs_model_deployer']);
+        new Function(this, functionId, {
+            functionName: functionId,
+            runtime: Runtime.NODEJS_18_X,
+            handler: 'index.handler',
+            code: Code.fromAsset('./ecs_model_deployer/dist/'),
+            timeout: Duration.minutes(10),
```
petermuller, line 57: Since it seems like we're going to wait for the whole model deployment to happen, should we just give the maximum wait time of 15 minutes?

ecs_model_deployer/src/lib/app.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const configFile = JSON.parse(process.env['LISA_CONFIG']!) as ConfigFile;
+const config = ConfigSchema.parse(configFile);
+
+const modelConfig = JSON.parse(process.env['LISA_MODEL_CONFIG']!);
+
+const stackProps: LisaModelStackProps = {
+    env: {
+        account: process.env['CDK_DEFAULT_ACCOUNT'],
+        region: process.env['CDK_DEFAULT_REGION']
```
petermuller, line 32: are these always guaranteed to be set?

lib/models/ecs-model-deployer.ts
stephensmith-aws: We're actually not waiting for the entire stack creation/destruction. Right now I have it configured to wait 1 minute after starting the CDK command, then exit. In my testing, this has been enough time for the `create-stack` operation to hit cloudformation, so we can safely exit without affecting the actual deployment. I wanted to poll against `describe-stack` but unfortunately adding `@aws-sdk/client-cloudformation` pushed us over the 250MB limit.

ecs_model_deployer/src/lib/app.ts
stephensmith-aws: I believe so based on [this page](https://docs.aws.amazon.com/cdk/v2/guide/configure-env.html).
> The AWS CDK provides two environment variables that you can use within your CDK code: CDK_DEFAULT_ACCOUNT and CDK_DEFAULT_REGION. When you use these environment variables within the env property of your stack instance, you can pass environment information from your credentials and configuration files using the CDK CLI --profile option.

ecs_model_deployer/.gitignore
stephensmith-aws: If we wanted to get rid of this, we'd need to do `ecs_model_deployer/dist`, or `**/dist`. The former would break if we renamed the directory, the latter feels a little dangerous for a monorepo, which is why I added the nested `.gitignore`. If you prefer one of the solutions from here though that would also totally work so just let me know and I'll update it.

ecs_model_deployer/src/index.ts
stephensmith-aws: good catch, somehow left off the `else { throw err; }`. The intention of this was to eat the EEXIST error but let anything else throw. Will update this.

ecs_model_deployer/src/lib/app.ts
petermuller: From the same page, I'm gathering that these are still something that we'd have to intentionally define (otherwise how does it know what to set?)- this statement toward the bottom of the page makes me think that they aren't guaranteed to exist:
> These environment variables can override defaults at synthesis time if they exist

But I'm also assuming it worked in your account though, right? So maybe a non-issue?

lib/models/ecs-model-deployer.ts
petermuller: Ah I see, not a problem. That seems fine then. 10 minutes may be a lot, but if we're not at risk of the lambda cutting off before the operations have completed, then I'm good.

ecs_model_deployer/src/lib/app.ts
stephensmith-aws: I believe the "if they exist" there is referring to the non-CDK provided `CDK_DEPLOY_ACCOUNT` and `CDK_DEPLOY_REGION`, and then the fallback is to used the CDK provided ones that I'm using. Yeah, it is working in my account, and my understanding is that it just gets set based on the account your `AWS_*` variables point to.

ecs_model_deployer/.gitignore
petermuller: all good- i haven't seen a nested gitignore like that before 😅

petermuller (commented):
And just to confirm, after deploying your changes, were you able to:
1. validate that you were still able to query models through direct API calls with the ALB?
2. run the chat UI
    1. have chat messages returned in the window
    2. perform rag operations with an embedding model and datastore
    3. not have any 5xx errors in the network console when clicking around in the UI?

stephensmith-aws:
> And just to confirm, after deploying your changes, were you able to:
> 
>     1. validate that you were still able to query models through direct API calls with the ALB?
> 
>     2. run the chat UI
>        
>        1. have chat messages returned in the window
>        2. perform rag operations with an embedding model and datastore
>        3. not have any 5xx errors in the network console when clicking around in the UI?

1. Yes
2. I haven't done anything in the UI. My understanding is that since we don't have the CreateModel workflow actually working and nothing is talking to LiteLLM the UI doesn't know about my models

petermuller:
I think things are looking ok- Just need to address the precommit stuff then I'll be all set. You can run those locally if you have your python environment activated, then you just have to run:

```
pre-commit run --all-files && npm test
```

you might need to pip install something else in your env, but this is the shortcut i use in my ide to avoid the guesswork

petermuller (approved):

PR #57, opened 2024-08-28
Implement ListModels, GetModel, DeleteModel model management APIs

petermuller:
This set of changes implements the ListModels, GetModel, and DeleteModel API for the model management feature. 
* ListModels will list all the models that are stored in the LiteLLM database and return any information that we store as additional metadata through the Create API once that one is implemented.
* GetModel will get the metadata for one specific model. It appears LiteLLM is unable to filter for the model on its own, so this is implemented as a filter over the ListModels API.
* DeleteModel in its current form will delete a reference from LiteLLM. Once we have more information from the CreateModel API, we can add logic to delete the CloudFormation stack associated with the custom infrastructure we spin up for users in this API.

Main changes
* Addition of LiteLLM client to make the management API calls directly
* Creation of a handler class for each specific API- this will start a pattern of offloading the handler logic outside of the REST API definitions file to keep the API definition file a little cleaner
* Abstracted the base handler class to make class definitions less repetitive to implement
* Added functionality for ListModels, GetModel
* Created initial revision of DeleteModel

### DIFF
--- a/lambda/models/clients/__init__.py
+++ b/lambda/models/clients/__init__.py
@@ -1,0 +1,13 @@
+#   Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+#   Licensed under the Apache License, Version 2.0 (the "License").
+#   You may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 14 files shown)


petermuller:
Looking into the precommit failure- didn't see that when I ran locally.

estohlmann (approved):

PR #61, opened 2024-08-29
Updating UI cards and added ModelURL to API call for create

estohlmann:
*Description of changes:*

Updated Model cards to display Hosted Platform and Instance type. Also update create API call to pass in ModelURL so that the call will complete.

Screenshot 2024-08-29 at 3 53 09 PM

### DIFF
--- a/lib/user-interface/react/src/components/model-management/ModelManagementUtils.tsx
+++ b/lib/user-interface/react/src/components/model-management/ModelManagementUtils.tsx
@@ -56,8 +56,8 @@ export const CARD_DEFINITIONS = {
         },
         {
             id: 'Hosting',
-            header: 'Hosting Platform',
-            content: (model: IModel) => (model.ContainerConfig === null && model.AutoScalingConfig === null && model.LoadBalancerConfig === null) ? 'External Provider' : 'LISA Hosted',
+            header: 'Hosted in LISA',
+            content: (model: IModel) => String(model.ContainerConfig === null && model.AutoScalingConfig === null && model.LoadBalancerConfig === null),
         },
         {
             id: 'InstanceType',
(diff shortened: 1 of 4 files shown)


lib/user-interface/react/src/components/model-management/ModelManagementUtils.tsx
```diff
@@ -47,13 +47,23 @@ export const CARD_DEFINITIONS = {
...
         },
         {
             id: 'Streaming',
             header: 'Streaming',
             content: (model: IModel) => String(model.Streaming),
         },
+        {
+            id: 'Hosting',
+            header: 'Hosting Platform',
+            content: (model: IModel) => (model.ContainerConfig === null && model.AutoScalingConfig === null && model.LoadBalancerConfig === null) ? 'LiteLLM' : 'LISA Hosted',
```
petermuller, line 60: All of these are being processed by LiteLLM, whether we host the infrastructure or not. For our own infrastructure we just use "OpenAI" configurations with a url to our models' load balancers, and for anything else, that can be a reference to another cloud provider, or ollama, or sagemaker, bedrock, etc. And all those configurations need to be inside LiteLLM in order for the models to resolve for users.

I think this distinction makes the config a little confusing, but I understand what we're going toward. I'd suggest something more like: `LISA Hosted` and maybe `Custom Provider` to make those distinctions. Also open to other names as I'm failing to think of better names to use

estohlmann: How about `External Provider`

petermuller: Bedrock and SageMaker are included in this- I'd not really consider them "external" since it's deploying in an AWS account

estohlmann: Per slack thread changed field to boolean 'Hosted in LISA'

petermuller (approved):

PR #65, opened 2024-09-05
DockerImageBuild state machine implementation

stephensmith-aws:
*Description of changes:*

Implement docker image build piece of state machine

### DIFF
--- a/lambda/dockerimagebuilder/__init__.py
+++ b/lambda/dockerimagebuilder/__init__.py
@@ -35,16 +35,17 @@ docker push {{ECR_URI}}:{{IMAGE_ID}}
 """
 
 
-def handler(event: Dict[str, Any], context) -> None:  # type: ignore [no-untyped-def]
+def handler(event: Dict[str, Any], context) -> Dict[str, Any]:  # type: ignore [no-untyped-def]
     base_image = event["base_image"]
     layer_to_add = event["layer_to_add"]
-    mounts3_deb_url = event["mounts3_deb_url"]
+    mounts3_deb_url = os.environ["LISA_MOUNTS3_DEB_URL"]
 
     ec2_resource = boto3.resource("ec2", region_name=os.environ["AWS_REGION"])
     ssm_client = boto3.client("ssm", region_name=os.environ["AWS_REGION"])
 
     response = ssm_client.get_parameter(Name="/aws/service/ami-amazon-linux-latest/amzn2-ami-hvm-x86_64-gp2")
     ami_id = response["Parameter"]["Value"]
+    image_tag = str(uuid.uuid4())
 
     rendered_userdata = user_data_template
     rendered_userdata = rendered_userdata.replace("{{AWS_REGION}}", shlex.quote(os.environ["AWS_REGION"]))
@@ -53,12 +54,10 @@ def handler(event: Dict[str, Any], context) -> None:  # type: ignore [no-untyped
     rendered_userdata = rendered_userdata.replace("{{BASE_IMAGE}}", shlex.quote(base_image))
     rendered_userdata = rendered_userdata.replace("{{MOUNTS3_DEB_URL}}", shlex.quote(mounts3_deb_url))
     rendered_userdata = rendered_userdata.replace("{{ECR_URI}}", shlex.quote(os.environ["LISA_ECR_URI"]))
-    rendered_userdata = rendered_userdata.replace("{{IMAGE_ID}}", str(uuid.uuid4()))
+    rendered_userdata = rendered_userdata.replace("{{IMAGE_ID}}", image_tag)
 
-    print("Creating instance with userdata: ")
-    print(rendered_userdata)
     try:
-        ec2_resource.create_instances(
+        instances = ec2_resource.create_instances(
             ImageId=ami_id,
             MinCount=1,
             MaxCount=1,
@@ -66,7 +65,16 @@ def handler(event: Dict[str, Any], context) -> None:  # type: ignore [no-untyped
             UserData=rendered_userdata,
             IamInstanceProfile={"Arn": os.environ["LISA_INSTANCE_PROFILE"]},
             BlockDeviceMappings=[{"DeviceName": "/dev/xvda", "Ebs": {"VolumeSize": 32}}],
+            TagSpecifications=[
+                {
+                    "ResourceType": "instance",
+                    "Tags": [
+                        {"Key": "Name", "Value": f"LisaDockerImageBuilder_{image_tag}"},
+                        {"Key": "lisa_temporary_instance", "Value": "true"},
+                    ],
+                }
+            ],
         )
+        return {"instance_id": instances[0].instance_id, "image_tag": image_tag}
     except ClientError as e:
-        print(e)
-        pass
+        raise e
--- a/lambda/models/state_machine/create_model.py
+++ b/lambda/models/state_machine/create_model.py
@@ -14,27 +14,83 @@
 
 """Lambda handlers for CreateModel state machine."""
 
+import json
+import os
 from copy import deepcopy
 from typing import Any, Dict
+
+import boto3
+from models.domain_objects import CreateModelRequest
+
+lambdaClient = boto3.client("lambda")
+ecrClient = boto3.client("ecr")
+ec2Client = boto3.client("ec2")
+ddbResource = boto3.resource("dynamodb")
+model_table = ddbResource.Table(os.environ["MODEL_TABLE_NAME"])
 
 
 def handle_set_model_to_creating(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
     """Set DDB entry to CREATING status."""
     output_dict = deepcopy(event)
-    output_dict["create_infra"] = True
+    request = CreateModelRequest.validate(event)
+
+    if (
+        request.AutoScalingConfig  # noqa: W503
+        and request.ContainerConfig  # noqa: W503
+        and request.InferenceContainer  # noqa: W503
+        and request.InstanceType  # noqa: W503
+        and request.LoadBalancerConfig  # noqa: W503
+    ):
+        model_table.update_item(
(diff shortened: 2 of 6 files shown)


lambda/models/state_machine/create_model.py
```diff
@@ -16,25 +16,62 @@
...
+
+import boto3
+from botocore.exceptions import ClientError
 
 
 def handle_set_model_to_creating(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
     """Set DDB entry to CREATING status."""
     output_dict = deepcopy(event)
-    output_dict["create_infra"] = True
+    output_dict["create_infra"] = 'modelConfig' in event
```
petermuller, line 29: Ah I see what you mean now- for these, we can use the names as they're defined in the CreateModelRequest, which we can pass as direct input into the state machine. I would suggest a few things here:

1. Let's define any aws clients at the top of the file, similar to how they're done here: https://github.com/awslabs/LISA/blob/283dd8f4a2be3288e0e251a41e838e75e807fef0/lambda/models/state_machine/delete_model.py#L28-L30 (these variables are already set from the CDK definition- we can add more variables as needed)
2. For this create_infra variable, let's make it be `True` if all of AutoScalingConfig, ContainerConfig, InferenceContainer, InstanceType, and LoadBalancerConfig are not `None`.  For False, all of these need to be None, and then let's throw an exception if they're partially configured (some null, some not). Basically we have two use cases:
    1. We define the ECS infrastructure which is all of those fields, and ModelUrl must also be none
    2. We define a LiteLLM-only entry, which requires all of the others to be null, and we'd set the ModelUrl as specified in the request. For the initial revision here, let's just worry about the first case.
3.  In this step specifically, let's make sure that we set the model state to `Creating` in DDB, and we can add other details, like last modified date, the entire request blob, and anything else that seems like it'll be useful to have around. We can also add all that to the output_dict so that future states can have that info without calling out to ddb

lib/models/ecs-model-deployer.ts
```diff
@@ -29,6 +29,7 @@ export type ECSModelDeployerProps = {
 } & BaseProps;
 
 export class ECSModelDeployer extends Construct {
+    readonly fn: Function;
```
petermuller, line 32: let's make this name more descriptive so that we can identify what it is, something like "ecsModelDeployerFunction" or similar, that way if we need to grab another item out of this construct in the future, we have less have a chance of ambiguously named resources. (and we'd need to change any of the references that use it too)

lambda/dockerimagebuilder/__init__.py
```diff
@@ -53,20 +54,38 @@ def handler(event: Dict[str, Any], context) -> None:  # type: ignore [no-untyped
...
     try:
-        ec2_resource.create_instances(
+        instances = ec2_resource.create_instances(
             ImageId=ami_id,
             MinCount=1,
             MaxCount=1,
             InstanceType="m5.large",
             UserData=rendered_userdata,
             IamInstanceProfile={"Arn": os.environ["LISA_INSTANCE_PROFILE"]},
             BlockDeviceMappings=[{"DeviceName": "/dev/xvda", "Ebs": {"VolumeSize": 32}}],
```
petermuller, line 69: does the block device get deleted on termination too? just want to confirm that while we're here

lambda/dockerimagebuilder/__init__.py
```diff
@@ -53,20 +54,38 @@ def handler(event: Dict[str, Any], context) -> None:  # type: ignore [no-untyped
...
+                        }
+                    ]
+                }
+            ]
         )
+        return {
+            "instance_id": instances[0].instance_id,
+            "image_tag": image_tag
+        }
     except ClientError as e:
```
petermuller, line 90: I'm not sure which errors we're expecting, but I think we can just let them boil out to the state machine. From the state machine definition after it's deployed, it automatically handles some form of sdk errors, but I'm not 100% sure if those are strictly lambda or if they're any sdk call:

```json
      "Retry": [
        {
          "ErrorEquals": [
            "Lambda.ClientExecutionTimeoutException",
            "Lambda.ServiceException",
            "Lambda.AWSLambdaException",
            "Lambda.SdkClientException"
          ],
          "IntervalSeconds": 2,
          "MaxAttempts": 6,
          "BackoffRate": 2
        }
      ],
```

lambda/models/state_machine/create_model.py
```diff
@@ -16,25 +16,62 @@
...
     return output_dict
 
 
 def handle_start_copy_docker_image(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
     """Start process for copying Docker image into local AWS account."""
     output_dict = deepcopy(event)
+
+    lambdaClient = boto3.client("lambda")
```
petermuller, line 37: same note from my other comment, let's add any clients to the top of the file to take advantage of "warm" lambda functions that can reuse the clients defined at the module level

lambda/models/state_machine/create_model.py
```diff
@@ -16,25 +16,62 @@
...
+    print(f"Got payload {payload}")
+    output_dict["image_info"] = json.loads(payload)
     return output_dict
 
 
 def handle_poll_docker_image_available(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
     """Check that Docker image is available in account or not."""
     output_dict = deepcopy(event)
+
+    ecrClient = boto3.client("ecr")
```
petermuller, line 54: same note for clients, let's define them at the top of the file with the region and retry config

lambda/models/state_machine/create_model.py
```diff
@@ -16,25 +16,62 @@
...
+    try:
+        response = ecrClient.describe_images(
+            repositoryName=os.environ["ECR_REPOSITORY_NAME"],
+            imageIds=[
+                {
+                    "imageTag": event["image_info"]["image_tag"]
+                }
+            ]
+        )
+    except ClientError as e:
```
petermuller, line 64: do we get anything more specific than  a "ClientError" if the image doesn't exist yet? I'd prefer to have the exact exception so we don't unintentionally catch errors that we shouldn't attempt to retry on

lambda/models/state_machine/create_model.py
```diff
@@ -16,25 +16,62 @@
...
+        output_dict["continue_polling_docker"] = True
+        return output_dict 
+
     output_dict["continue_polling_docker"] = False
+    ec2Client = boto3.client("ec2")
```
petermuller, line 69: same note on client definitions at the top of the file :)

lambda/dockerimagebuilder/__init__.py
stephensmith-aws: yep just confirmed

lambda/models/state_machine/create_model.py
stephensmith-aws: narrowed down scope to ImageDoesNotExist

.pre-commit-config.yaml
```diff
@@ -80,7 +80,7 @@ repos:
         - --docstring-convention=numpy
         - --max-line-length=120
         - --extend-immutable-calls=Query,fastapi.Depends,fastapi.params.Depends
-        - --ignore=B008 # Ignore error for function calls in argument defaults
+        - --ignore=B008,W503 # Ignore error for function calls in argument defaults
```
petermuller, line 83: I think we should probably keep W503 to keep with the pep standards https://www.flake8rules.com/rules/W503.html

where did this happen?

stephensmith-aws: happened in create model state machine implementation. One of our formatters wants to put each condition on its own line, and flake8 doesn't. Couldn't make them both happy at the same time.
`lambda/models/state-machine/create-model.py:38`

petermuller: Oh for that we could try putting the `and`'s at the end of the line.  i don't think there's anything wrong with needing the backslash at the end to complete the line either. otherwise we can just do an inline ignore so we ignore the one line, but not the rest. so adding a comment like `# noqa: W503` on the specific line could work

lambda/models/state_machine/create_model.py
```diff
@@ -14,27 +14,77 @@
...
+
 
 def handle_set_model_to_creating(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
     """Set DDB entry to CREATING status."""
     output_dict = deepcopy(event)
-    output_dict["create_infra"] = True
+    request = CreateModelRequest.validate(event)
+
+    if (
+        request.AutoScalingConfig
```
petermuller, line 39: depending on the formatting, we can try to put the `and` at the end of each line with a continuation slash if needed, or we can put a comment to have the formatter ignore this situation specifically `#noqa: W503` or something like that

stephensmith-aws: no matter how I format it, black will change the formatting to what I currently have, then flake8 will come along and fail afterwards

stephensmith-aws: changed to use inline ignores instead of global

lambda/models/state_machine/create_model.py
```diff
@@ -14,27 +14,76 @@
...
+
+    payload = response["Payload"].read()
+    output_dict["image_info"] = json.loads(payload)
     return output_dict
 
 
 def handle_poll_docker_image_available(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
     """Check that Docker image is available in account or not."""
     output_dict = deepcopy(event)
+
```
petermuller, line 76: Talked about over chat, but let's also add a field in the event/output_dict to hold a "poll_docker_iterations" field that can start at 0 and then get incremented in each iteration. once we exceed a certain amount of time (we wait a minute between polls), then we can just decide to stop iterating, clean up the ec2 instance, and raise an exception to fail the state machine

stephensmith-aws: added

petermuller (approved):

PR #66, opened 2024-09-05
Moving ALB behind APIGW to allow service code to call apis without auth

estohlmann:


lib/api-base/fastApiContainer.ts
```diff
@@ -109,16 +112,58 @@ export class FastApiContainer extends Construct {
...
                 identifier: props.apiName,
                 instanceType: taskConfig.instanceType,
-                internetFacing: config.restApiConfig.internetFacing,
+                internetFacing: false,
                 loadBalancerConfig: taskConfig.loadBalancerConfig,
             },
             securityGroup,
             vpc,
+            addNlb: true
+        });
```
petermuller, line 121: Reading through the resources, it looks like ECSCluster->alb (apiCluster.alb) will ultimately be the load balancer that serves as the direct (unauthenticated) entrypoint to the EC2 instances that are running the ECS services for the Rest API. Is that correct?

If so, can we also add the load balancer (alb) as an externally accessible property on this construct (add to the "public readonly" entries at the top)? this would help us to stick the load balancer address into the models api and rag stacks (most likely as an environment variable) so that we can still use the endpoint for any model operations (and that would have to bubble up to stages.ts for that to get passed into those stacks)

lib/user-interface/index.ts
```diff
@@ -187,11 +185,7 @@ export class UserInterfaceStack extends Stack {
             ADMIN_GROUP: config.authConfig!.adminGroup,
             JWT_GROUPS_PROP: config.authConfig!.jwtGroupsProperty,
             CUSTOM_SCOPES: config.authConfig!.additionalScopes,
-            RESTAPI_URI: StringParameter.fromStringParameterName(
```
petermuller, line 190: this is removed- but have we removed the references that depend on it?

estohlmann: It is defined down below

lib/api-base/fastApiContainer.ts
petermuller: I just realized I'm using a SSM parameter to get the lisa serve address anyways. if that's still there then we don't need to change anything

estohlmann: I don't think I changed anything there but I added it to the exposed readonly props as well :D

lib/api-base/fastApiContainer.ts
```diff
@@ -57,6 +70,8 @@ export class FastApiContainer extends Construct {
     /** FastAPI URL **/
     public readonly endpoint: string;
 
+    public readonly alb: ApplicationLoadBalancer;
```
petermuller, line 73: my bad- I missed the fact that we pass back the "endpoint" which is actually the ALB address 😬 
we can remove lol

For my purposes, I just need the "endpoint", and searching around for that, RAG is already taken care of (so long as that's also in the VPC), and I just need to physically add it to the models stack. sorry for the churn on that.

lib/api-base/fastApiContainer.ts
```diff
@@ -109,16 +115,60 @@ export class FastApiContainer extends Construct {
...
+        resource.addCorsPreflight({
+            allowOrigins: Cors.ALL_ORIGINS,
+            allowHeaders: ['*'],
         });
 
         if (tokenTable) {
             tokenTable.grantReadData(apiCluster.taskRole);
         }
+
         this.endpoint = apiCluster.endpointUrl;
```
petermuller, line 172: I missed this detail- looking in the cluster definition, this is already the ALB address

lib/api-base/fastApiContainer.ts
estohlmann: No worries

petermuller (approved):

petermuller (approved):

PR #71, opened 2024-09-06
create/poll stack implementations

stephensmith-aws:
create/poll stack implementations

9/6 0300: still need to test polling, there might be an issue with long images getting built connection reset by peer. issue with environment variables getting camelized then not making it to container

9/6 0630: Got a completely zero-touch create model execution that built the image, started the stack, polled it successfully until it completed.

### DIFF
--- a/ecs_model_deployer/src/index.ts
+++ b/ecs_model_deployer/src/index.ts
@@ -17,9 +17,6 @@
 import { spawnSync, spawn, ChildProcess } from 'child_process';
 import { readdirSync, symlinkSync, rmSync } from 'fs';
 
-const ACTION_DEPLOY = 'deploy';
-const ACTION_DESTROY = 'destroy';
-
 /*
   cdk CLI always wants ./ to be writable in order to write cdk.context.json.
   This should really be an environment variable or something, but this function
@@ -54,17 +51,9 @@ const createWritableEnv = () => {
 };
 
 export const handler = async (event: any) => {
-    if (!event.action) {
-        console.log(`action not provided in ${JSON.stringify(event)}`);
-        throw new Error('action not provided');
-    } else if ( ![ACTION_DESTROY, ACTION_DEPLOY].includes(event.action) ) {
-        console.log(`Invalid action ${event.action}`);
-        throw new Error(`Invalid action ${event.action}`);
-    }
-
     if (!event.modelConfig) {
         console.log(`modelConfig not provided in ${JSON.stringify(event)}`);
-        throw new Error('modeConfig not provided');
+        throw new Error('modelConfig not provided');
     }
     const modelConfig = event.modelConfig;
     process.env['LISA_MODEL_CONFIG'] = JSON.stringify(modelConfig);
@@ -79,8 +68,7 @@ export const handler = async (event: any) => {
 
     const ret = spawnSync('./node_modules/aws-cdk/bin/cdk', ['synth', '-o', '/tmp/cdk.out']);
 
-    let stdout = String(ret.output[1]);
-    let stderr = String(ret.output[2]);
+    const stderr = String(ret.output[2]);
     if ( ret.status !== 0 ) {
         console.log(`cdk synth failed with stderr: ${stderr}`);
         throw new Error('Stack failed to synthesize');
@@ -88,56 +76,34 @@ export const handler = async (event: any) => {
 
 
     const stackName = `${config.deploymentName}-${modelConfig.modelId}`;
-    if ( event.action === ACTION_DEPLOY ) {
-        const deploy_promise: Promise<ChildProcess | undefined> = new Promise( (resolve) => {
-            const cp = spawn('./node_modules/aws-cdk/bin/cdk', ['deploy', stackName, '-o', '/tmp/cdk.out']);
+    const deploy_promise: Promise<ChildProcess | undefined> = new Promise( (resolve) => {
+        const cp = spawn('./node_modules/aws-cdk/bin/cdk', ['deploy', stackName, '-o', '/tmp/cdk.out']);
 
-            cp.on('close', (code) => {
-                console.log(`cdk deploy exited early, code ${code}`);
-                resolve(cp);
-            });
-
-            cp.stdout.on('data', (data) => {
-                console.log(`Got data: ${data}`);
-            });
-
-            cp.stderr.on('data', (data) => {
-                console.log(`Got err data: ${data}`);
-            });
-
-            setTimeout(() => {
-                console.log('180 second timeout');
-                resolve(undefined);
-            }, 180 * 1000);
+        cp.on('close', (code) => {
+            console.log(`cdk deploy exited early, code ${code}`);
+            resolve(cp);
         });
 
-        const cp = await deploy_promise;
-        if ( cp ) {
-            if ( cp.exitCode !== 0 ) {
-                throw new Error('Stack failed to deploy');
-            }
-        }
-    } else if ( event.action === ACTION_DESTROY ) {
-        const deploy_promise: Promise<Number> = new Promise( (resolve) => {
-            const cp = spawn('./node_modules/aws-cdk/bin/cdk', ['destroy', '-f', stackName, '-o', '/tmp/cdk.out']);
-
-            cp.on('close', (code) => {
-                resolve(code ?? -1);
-            });
-
-            setTimeout(() => {
-                console.log('60 second timeout');
-                resolve(0);
-            }, 180 * 1000);
+        cp.stdout.on('data', (data) => {
+            console.log(`Got data: ${data}`);
         });
 
-        const exitCode = await deploy_promise;
-        stdout = String(ret.output[1]);
-        stderr = String(ret.output[2]);
-        if ( exitCode !== 0 ) {
-            console.log(`cdk destroy failed with stdout: ${stdout}, stderr: ${stderr}`);
-            throw new Error('Stack failed to destroy');
+        cp.stderr.on('data', (data) => {
+            console.log(`Got err data: ${data}`);
+        });
+
+        setTimeout(() => {
+            console.log('180 second timeout');
+            resolve(undefined);
+        }, 180 * 1000);
+    });
+
+    const cp = await deploy_promise;
+    if ( cp ) {
+        if ( cp.exitCode !== 0 ) {
+            throw new Error('Stack failed to deploy');
         }
     }
-    return stackName;
+
+    return {stackName: stackName};
 };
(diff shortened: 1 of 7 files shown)


lambda/models/state_machine/create_model.py
```diff
@@ -97,14 +99,61 @@ def handle_poll_docker_image_available(event: Dict[str, Any], context: Any) -> D
 def handle_start_create_stack(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
     """Start model infrastructure creation."""
     output_dict = deepcopy(event)
+    request = CreateModelRequest.validate(event)
+
+    prepared_event = {}
+
+    def camelize_object(o):
```
petermuller, line 106: this is a lot of work to transform an object that we control ourselves. aside from the input object, can we create data in the format that we expect later in the state machine or in subsequent invocations? I'm also good with just embedding the entire raw request inside the event object we pass along so that we can reference anything we want in it

lambda/models/state_machine/create_model.py
```diff
@@ -97,14 +99,61 @@ def handle_poll_docker_image_available(event: Dict[str, Any], context: Any) -> D
...
+
+    response = lambdaClient.invoke(
+        FunctionName=os.environ['ECS_MODEL_DEPLOYER_FN_ARN'],
+        Payload=json.dumps({"modelConfig": prepared_event}),
+    )
+
+    payload = response["Payload"].read()
+    payload = json.loads(payload)
+    stack_name = payload.get('stackName')
+    output_dict["stack_name"] = stack_name
```
petermuller, line 131: could we change this to a stack arn instead? the whole arn with the trailing UUID will help us identify the stack, even if it is ultimately rolled back and deleted before we're done monitoring it, otherwise if it gets deleted and we don't have its uuid, we may not be able to track it after a while

example for a stack whose name is `my-model-stack`:

```
arn:aws:cloudformation:us-east-1:000000000000:stack/my-model-stack/dafea8c0-a96b-11ec-af1e-1285d9cd23a9
```

lambda/models/state_machine/create_model.py
```diff
@@ -97,14 +99,61 @@ def handle_poll_docker_image_available(event: Dict[str, Any], context: Any) -> D
...
+    model_table.update_item(
+        Key={"model_id": request.ModelId}, AttributeUpdates={"StackName": {"Value": stack_name, "Action": "PUT"}}
```
petermuller, line 133: AttributeUpdates is deprecated. Can we migrate to UpdateExpression instead?

https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/LegacyConditionalParameters.AttributeUpdates.html

https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/LegacyConditionalParameters.AttributeUpdates.html#UpdateExpression.instead


I think I may have an example of what that would look like in the DeleteModel workflow

lambda/models/state_machine/create_model.py
```diff
@@ -97,14 +99,61 @@ def handle_poll_docker_image_available(event: Dict[str, Any], context: Any) -> D
...
+    )
+
+    output_dict["remaining_polls_stack"] = 15
```
petermuller, line 136: I can picture the models to take a while, depending on how large they are. I'd possibly suggest bumping up to 30, but it may be ok as is since at this point the time is strictly on instance startup and doesn't include the time for building the docker image.

ecs_model_deployer/src/index.ts
```diff
@@ -88,56 +80,34 @@ export const handler = async (event: any) => {
...
-            cp.stderr.on('data', (data) => {
-                console.log(`Got err data: ${data}`);
-            });
-
-            setTimeout(() => {
-                console.log('180 second timeout');
-                resolve(undefined);
-            }, 180 * 1000);
+    const deploy_promise: Promise<ChildProcess | undefined> = new Promise( (resolve) => {
+        const cp = spawn('./node_modules/aws-cdk/bin/cdk', ['deploy', stackName, '-o', '/tmp/cdk.out']);
```
petermuller, line 84: I know we're using cdk here, but is it also possible to set a Role to be associated with the cloudformation stack? (if it doesn't do that already?)

DeleteModel doesn't have the Delete* permissions that would be required, but that's ok if the stack has an associated role that already has those permissions.

ecs_model_deployer/src/index.ts
```diff
@@ -88,56 +80,34 @@ export const handler = async (event: any) => {
...
-            throw new Error('Stack failed to destroy');
+        cp.stderr.on('data', (data) => {
+            console.log(`Got err data: ${data}`);
+        });
+
+        setTimeout(() => {
+            console.log('180 second timeout');
+            resolve(undefined);
+        }, 180 * 1000);
+    });
```
petermuller, line 103: What's going on in this block? 3 minute wait? Are we expecting this lambda to stay running the entire duration of the cdk update?

lambda/models/state_machine/create_model.py
stephensmith-aws: not sure I'm understanding the "embedding the entire raw request" piece here. The goal of this is to fix all of our PascalPropertyNames that we have from pydantic into camelPropertyNames that the CDK schema expects. We could (and probably should) go through and change them all to pascal case but I just did this in the interest of moving quickly.

lambda/models/state_machine/create_model.py
stephensmith-aws: yeah, worst case I can just do a `describeStacks` by name right after the lambda returns. Not sure if there's going to be a way to get the stack arn out of cdk before it finishes deploying

lambda/models/state_machine/create_model.py
stephensmith-aws: will do

ecs_model_deployer/src/index.ts
stephensmith-aws: We need to wait long enough to be confident that cdk has under the hood called `aws cloudformation create-stack`. If we exit before this time, the deployment will not happen. We are essentially timing a "control-c" to `cdk deploy` to be late enough that it doesn't matter. I wanted to poll on cloudformation describestacks but unfortunately @aws-sdk/client-cloudformation took us over 250MB...

stephensmith-aws: but if CDK fails within the 3 minutes (like a synth fail) we will still capture that error.

ecs_model_deployer/src/index.ts
stephensmith-aws: To be honest I'm not I understand the question. I will say for the cdk function, the lambda has 'sts:AssumeRole' on 'cdk*' roles, and it assumes that role before deploying anything.

I also used to have the 'action' parameter that would let us call `cdk destroy` instead of `aws cloudformation delete-stack` directly, so if we run into permissions issues we could move back to that and have cdk just assume the right role automatically. The big drawback of this is that you would need to pass in the entire modelConfig again in order to synth

petermuller: the equivalent of: can we set the `--role-arn` property on the stack through cdk? https://docs.aws.amazon.com/cli/latest/reference/cloudformation/create-stack.html

lambda/models/state_machine/create_model.py
petermuller: yea that works- i'm good with describe to get the arn. the arn will be most consistent to use in the rest of the state machine.

ecs_model_deployer/src/index.ts
petermuller: got it- so the 3 minutes is really just a brute-force "really make sure that the stack has started by this point" then? is it possible that we may not see that in 3 minutes? would it maybe make sense for us to make another wait+poll loop to see if the stack exists (since we know the name of the stack before starting the update)?

ecs_model_deployer/src/index.ts
stephensmith-aws: Looks like it defaults to cdk-cfn-exec-role... but that we can potentially change it. What would we want to change it to? If we give the Delete lambda 'sts:AssumeRole' on cdk* would that be enough?

ecs_model_deployer/src/index.ts
stephensmith-aws: the problem is we can't do that inside of the ecs_model_deployer lambda b/c of the 250MB limit. So we'd have to have it somewhere else in the state machine, which we essentially already do. The first time we hit the stack poller and it doesn't see that stack as CREATE_COMPLETE, CREATE_IN_PROGRESS, UPDATE_COMPLETE, UPDATE_IN_PROGRESS it will throw and abort the state machine with failure

petermuller: we're not locked into the current architecture- if we need to add one more polling loop in the state machine we can totally do that. I'm very ok with offloading that logic into another loop (outside of the large lambda func) so that we can track statuses appropriately and without missing an update.

i see it as:
loop 1: start docker copy -> poll docker loop
loop 2 (potential new loop): start cdk -> poll for stack creation started loop (don't even have to check status, just that the stack exists)
loop 3: check stack _success_ loop

ecs_model_deployer/src/index.ts
stephensmith-aws: Gotcha, the stack will have a role arn of something CDK created, so I think all we need then is cloudformation:DeleteStack on the deletemodel lambda role

lambda/models/state_machine/create_model.py
petermuller: got it, I misunderstood the "why" of doing this. all good 👍

lambda/models/state_machine/create_model.py
```diff
@@ -97,14 +100,69 @@ def handle_poll_docker_image_available(event: Dict[str, Any], context: Any) -> D
...
+        output_dict["continue_polling_stack"] = False
+        return output_dict
+    elif stackStatus in ["CREATE_IN_PROGRESS", "UPDATE_IN_PROGRESS"]:
+        output_dict["continue_polling_stack"] = True
+        output_dict["remaining_polls_stack"] -= 1
+        if output_dict["remaining_polls_stack"] <= 0:
+            raise Exception("Maximum number of CloudFormation polls reached")
+        return output_dict
+    else:
+        raise Exception("Stack in unexpected state.")
```
petermuller, line 165: Can we put the stack status in the error message too? Just to make it a little easier to see what state we're in? I think we may find some other states, but for now, I think these are ok to start- I would just like the status logged if there _is_ something awry

stephensmith-aws: good call, added

lambda/models/state_machine/.create_model.py.swp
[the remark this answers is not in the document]
stephensmith-aws: oh no, got caught being a psycho who uses vim.... I've done so well to not let any in until now

petermuller (approved):

PR #73, opened 2024-09-06
Add model data to ddb at beginning of state machine

stephensmith-aws:
Add model data to ddb at beginning of state machine

### DIFF
--- a/lambda/models/state_machine/create_model.py
+++ b/lambda/models/state_machine/create_model.py
@@ -45,7 +45,10 @@ def handle_set_model_to_creating(event: Dict[str, Any], context: Any) -> Dict[st
         and request.LoadBalancerConfig  # noqa: W503
     ):
         model_table.update_item(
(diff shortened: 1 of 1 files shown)


lambda/models/state_machine/create_model.py
```diff
@@ -45,7 +45,8 @@ def handle_set_model_to_creating(event: Dict[str, Any], context: Any) -> Dict[st
         and request.LoadBalancerConfig  # noqa: W503
     ):
         model_table.update_item(
-            Key={"model_id": request.ModelId}, AttributeUpdates={"Status": {"Value": "CREATING", "Action": "PUT"}}
+            Key={"model_id": request.ModelId},
+            AttributeUpdates={"Status": {"Value": "CREATING", "Action": "PUT"}, "ModelConfig": {"Value": event}},
```
petermuller, line 49: I missed this line in the UpdateExpressions conversion from earlier. Is that something that can be converted easily in the next few minutes? Otherwise, I'd say let's merge as-is, then we can refactor later

petermuller (approved):

PR #74, opened 2024-09-06
Documentation updates

petermuller:
This set of changes is a collection of the documentation updates we need to support the 3.0 launch.


petermuller:
As I'm focusing on the API implementations, I'll add some comments here of the information that I want to add:

## Breaking Changes
1. Existing deployed ECS/EC2-based models will be **deleted** if upgrading from v2 to v3. The model management UI or API can be used to recreate the models, but **it is expected that models infrastructure will be removed** as part of this migration. Static resources such as S3 artifacts or docker images will not be touched. This is simply the infrastructure as managed by CloudFormation that will be spun down.
1. Persistent API Tokens table will be deleted and replaced because we changed the name of it. The name change was required to allow for multiple LISA installations to happen in the same account. Please follow the instructions in the README file to populate the table with persistent tokens if desired. If you did not already use the table, then there is no additional action to take.
2. The RAG bucket will be recreated if you're updating from an older version of LISA. The bucket will need to be manually deleted if resources were uploaded to it already. CloudFormation will give details about which bucket is impacted. We recommend making a backup of documents in that bucket into another bucket before fully deleting it.
3. We require a key for the LiteLLM installation, which you may specify in the config.yaml file. This can be any arbitrary string that starts with 'sk-', and it will automatically apply when you do a LISA `cdk` deployment. The schema.ts file already enforces its existence.

petermuller:
## Differences in API definitions and config.yaml definitions
* Container filepaths
    * In config.yaml, file directory paths were relative to the repository root, so for vllm as an example, the path would be `lib/serve/ecs-model/vllm`, or for tgi, it would be `lib/serve/ecs-model/textgen/tgi` Now, with the model management API, directory paths are relative to $root/lib/serve/ecs-model, so the path would be `vllm` for vllm, or `textgen/tgi` for the tgi container

petermuller:
Addressed previous two comments in latest commit

estohlmann (approved):

PR #76, opened 2024-09-06
Apigw migration cleanup

estohlmann:
Removing un-needed code from the ALB Migration behind the API Gateway including:
- Auth code at the FastAPI layer
- Self Signed cert support
- Internet facing configuration for rest API
- Updating FE config to only reference one API endpoint

This PR also makes changes to the UI to:
- Make resource calls using ModelId instead of UniqueId
- Remove configs in create/update calls for externally hosted models 


README.md
```diff
@@ -155,7 +155,7 @@ permissions to the "REST-Role" that was created in the IAM stack:
 ```
 
 After adding those permissions and access in the VPC, LiteLLM will now be able to route traffic to those entities, and
-they will be accessible through the LISA ALB, using the OpenAI specification for programmatic access.
+they will be accessible through the LISA APIGateway, using the OpenAI specification for programmatic access.
```
petermuller, line 158: APIGateway -> API Gateway

README.md
```diff
@@ -546,33 +525,33 @@ routes as long as your underlying models can also respond to them.
 By supporting the OpenAI spec, we can more easily allow users to integrate their collection of models into their LLM applications and workflows. In LISA, users can authenticate
 using their OpenID Connect Identity Provider, or with an API token created through the DynamoDB token workflow as described [here](#programmatic-api-tokens). Once the token
 is retrieved, users can use that in direct requests to the LISA Serve REST API. If using the IdP, users must set the 'Authorization' header, otherwise if using the API token,
-users can set either the 'Api-Key' header or the 'Authorization' header. After that, requests to `https://${lisa_serve_alb}/v2/serve` will handle the OpenAI API calls. As an example, the following call can list all
-models that LISA is aware of, assuming usage of the API token. If you are using a self-signed cert, you must also provide the `--cacert $path` option to specify a CA bundle to trust for SSL verification.
+users can set either the 'Api-Key' header or the 'Authorization' header. After that, requests to `https://${lisa_api_gateway}/llm/v2/serve` will handle the OpenAI API calls. As an example, the following call can list all
```
petermuller, line 528: Do we want to remove the "Api-Key" support here?

README.md
```diff
@@ -492,9 +472,8 @@ window.env = {
   ADMIN_GROUP: '<The admin group you would like LISA to check the JWT token for>',
   CUSTOM_SCOPES:[<add your optional list of custom scopes to pull groups from your IdP here>],
   // Alternatively you can set this to be your REST api elb endpoint
-  RESTAPI_URI: 'http://localhost:8080/',
+  API_BASE_URL: '<API GW session endpoint>',
```
petermuller, line 475: Do we need to add a note to say that this also includes the stage name? maybe give an example of what that looks like?

```
https://${deployment_id}.execute-api.${regional_domain}/${deployment_stage}
```

README.md
```diff
@@ -600,31 +579,11 @@ client.models.list()
 
 To use the models being served by LISA, the client needs only a few changes:
 
-1. Specify the `base_url` as the LISA Serve ALB, using the /v2/serve route at the end, similar to the apiBase in the [Continue example](#continue-jetbrains-and-vs-code-plugin)
+1. Specify the `base_url` as the LISA API Gateway, using the /llm/v2/serve route at the end, similar to the apiBase in the [Continue example](#continue-jetbrains-and-vs-code-plugin)
 2. Add the API key that you generated from the [token generation steps](#programmatic-api-tokens) as your `api_key` field.
-3. If using a self-signed cert, you must provide a certificate path for validating SSL. If you're using an ACM or public cert, then this may be omitted.
-   1. We provide a convenience function in the `lisa-sdk` for generating a cert path from an IAM certificate ARN if one is provided in the `RESTAPI_SSL_CERT_ARN` environment variable.
 
 The Code block will now look like this and you can continue to use the library without any other modifications.
```
petermuller, line 585: we should still have a code block here right? we can remove the cert verify thing, but we just need the "base_url" defined with the https://apigw/dev/llm/v2/serve value, and then the api_key, and that's how LISA gets used

lambda/models/clients/litellm_client.py
```diff
@@ -25,11 +25,10 @@
 class LiteLLMClient:
     """Client definition for interfacing directly with LiteLLM management operations."""
 
-    def __init__(self, base_uri: str, headers: Headers, verify: Union[str, bool], timeout: int = 30):
```
petermuller, line 28: I'm working with these in my changes- if you want to drop these, i'll handle them in my followup changes. otherwise, we can get rid of the headers too since those won't be used anymore.

lib/serve/rest-api/src/api/routes.py
```diff
@@ -15,34 +15,22 @@
...
-router.include_router(models.router, prefix="/v1", tags=["models"], dependencies=dependencies, deprecated=True)
-router.include_router(embeddings.router, prefix="/v1", tags=["embeddings"], dependencies=dependencies, deprecated=True)
-router.include_router(generation.router, prefix="/v1", tags=["generation"], dependencies=dependencies, deprecated=True)
-router.include_router(
-    litellm_passthrough.router, prefix="/v2/serve", tags=["litellm_passthrough"], dependencies=dependencies
-)
+
+router.include_router(models.router, prefix="/v1", tags=["models"], deprecated=True)
+router.include_router(embeddings.router, prefix="/v1", tags=["embeddings"], deprecated=True)
+router.include_router(generation.router, prefix="/v1", tags=["generation"], deprecated=True)
```
petermuller, line 32: we've had these under the deprecation notice for a while now. this is already a large change, but maybe let's get rid of these and all associated code in a future commit if there's time. lets us remove more stuff from the readme.

lambda/models/clients/litellm_client.py
estohlmann: I think they should be removed since it is required

estohlmann: I will add them back and let you take card of it in your changes then

lib/serve/rest-api/src/api/routes.py
estohlmann: Maybe we can add that to the backlog for 3.0.1

README.md
```diff
@@ -347,9 +327,9 @@ pytest lisa-sdk/tests --url <rest-url-from-cdk-output> --verify <path-to-server.
 
 ## Programmatic API Tokens
 
-The LISA Serve ALB can be used for programmatic access outside the example Chat application.
+The LISA API Gateway can be used for programmatic access outside the example Chat application.
 An example use case would be for allowing LISA to serve LLM requests that originate from the [Continue VSCode Plugin](https://www.continue.dev/).
-To facilitate communication directly with the LISA Serve ALB, a user with sufficient DynamoDB PutItem permissions may add
+To facilitate communication directly with the LISA API Gateway, a user with sufficient DynamoDB PutItem permissions may add
 API keys to the APITokenTable, and once created, a user may make requests by including the `Authorization: Bearer ${token}`
 header or the `Api-Key: ${token}` header with that token. If using any OpenAI-compatible library, the `api_key` fields
```
petermuller, line 334: one more "Api-Key" reference we can remove; will approve after that

petermuller (approved):

petermuller (approved):

PR #80, opened 2024-09-07
UI API Updates

estohlmann:
Updating the UI to accept camel case for model management APIs, also fixing default values so that it will properly create externally hosted models.

Fixed a small bug around environment not properly being sent to BE.


petermuller (approved):
Looks good to me. From the changes I saw, it looks consistent with what I have on the backend

PR #81, opened 2024-09-07
Migrate APIs to DDB instead of LiteLLM db

petermuller:
This set of changes migrates from using the LiteLLM db to store state to the DynamoDB table that we're creating for this purpose. This allows us to store data before the model is fully finished, and lets us add any additional information that we'll need for debugging or investigation later.

Primary set of changes:
* API model refactor for camelcase to avoid Pydantic errors on truly optional parameters
* API handlers updated to accommodate the model refactor
* Relevant info now stored in DDB
* UI updates (from @estohlmann ) to work with API model update
* LiteLLM client no longer handles auth because traffic is all within VPC

### DIFF
--- a/lambda/models/clients/litellm_client.py
+++ b/lambda/models/clients/litellm_client.py
@@ -14,10 +14,9 @@
 
 """Client for interfacing with the LiteLLM proxy's management options directly."""
 
-from typing import Any, Dict, List, Union
+from typing import Any, Dict, List
 
 import requests
-from starlette.datastructures import Headers
 
 from ..exception import ModelNotFoundError
 
@@ -25,11 +24,9 @@ from ..exception import ModelNotFoundError
 class LiteLLMClient:
     """Client definition for interfacing directly with LiteLLM management operations."""
 
-    def __init__(self, base_uri: str, headers: Headers, verify: Union[str, bool], timeout: int = 30):
+    def __init__(self, base_uri: str, timeout: int = 30):
         self._base_uri = base_uri
-        self._headers = headers
         self._timeout = timeout
-        self._verify = verify
 
     def list_models(self) -> List[Dict[str, Any]]:
         """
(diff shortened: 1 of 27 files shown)


lambda/models/state_machine/create_model.py
```diff
@@ -171,4 +189,35 @@ def handle_poll_create_stack(event: Dict[str, Any], context: Any) -> Dict[str, A
...
+    output_dict["litellm_id"] = litellm_id
+
+    model_table.update_item(
+        Key={"model_id": event["modelId"]},
+        UpdateExpression="SET model_status = :ms, litellm_id = :lid, last_modified_date = :lm, model_url = :mu",
+        ExpressionAttributeValues={
+            ":ms": ModelStatus.IN_SERVICE,
+            ":lid": litellm_id,
+            ":lm": int(datetime.utcnow().timestamp()),
+            ":mu": litellm_params.get("api_base", ""),
```
petermuller, line 219: I think if I change this to None we can get consistent `null` in the UI if this is problematic

petermuller (commented):
I can't approve, but things looks good to me from here after looking through all the files.

lambda/models/state_machine/create_model.py
petermuller: Since it's already a string field, this didn't do anything- ddb will always respond with an empty string here.

lambda/models/clients/litellm_client.py
```diff
@@ -58,42 +54,38 @@ def add_model(self, model_name: str, litellm_params: Dict[str, str], additional_
...
         )
+        return resp.json()  # type: ignore [no-any-return]
 
-    def delete_model(self, unique_id: str) -> None:
+    def delete_model(self, identifier: str) -> None:
         """
         Delete a model from the database.
 
-        The unique_id is the ID that LiteLLM generates on its end when creating a model, regardless of if the model
+        The identifier is the ID that LiteLLM generates on its end when creating a model, regardless of if the model
```
estohlmann, line 71: Is it accurate that LiteLLM generates this?

lambda/models/lambda_functions.py
```diff
@@ -88,139 +83,131 @@ async def model_already_exists_handler(request: Request, exc: ModelAlreadyExists
...
-        headers=headers,
-        verify=get_cert_path(iam_client),
+        stepfunctions_client=stepfunctions,
+        model_table_resource=model_table,
+        litellm_client=litellm_client,
     )
     return list_handler()
 
 
 def _create_dummy_model(model_name: str, model_type: ModelType, model_status: ModelStatus) -> LISAModel:
```
estohlmann, line 108: I think we can delete this now right?

estohlmann (approved):

lambda/models/lambda_functions.py
petermuller: still used as placeholders for the update/start/stop api's at the moment

lambda/models/clients/litellm_client.py
petermuller: yes

estohlmann (approved):

PR #82, opened 2024-09-08
Disabling actions that won't be available with LISA 3.0.0

estohlmann:
Disabling the actions that won't be initially released


petermuller (commented):
Looks fine to me, but precommit failing because of doublequotes instead of single quotes for the `disabledReason`s

petermuller (approved):

PR #85, opened 2024-09-08
Updating to account for field validation

estohlmann:
Field validation to force required fields

### DIFF
--- a/lib/user-interface/react/src/components/model-management/create-model/CreateModelModal.tsx
+++ b/lib/user-interface/react/src/components/model-management/create-model/CreateModelModal.tsx
@@ -54,7 +54,11 @@ export function CreateModelModal (props: CreateModelModalProps) : ReactElement {
         updateModelMutation,
         { isSuccess: isUpdateSuccess, isError: isUpdateError, error: updateError, isLoading: isUpdating },
     ] = useUpdateModelMutation();
-    const initialForm = ModelRequestSchema.parse({});
+    const initialForm = {
+        ...ModelRequestSchema.parse({}),
+        modelId: '',
+        modelName: '',
+    };
     const dispatch = useAppDispatch();
     const notificationService = useNotificationService(dispatch);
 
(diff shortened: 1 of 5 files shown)


lib/user-interface/react/src/components/model-management/create-model/CreateModelModal.tsx
```diff
@@ -141,6 +145,8 @@ export function CreateModelModal (props: CreateModelModalProps) : ReactElement {
         }
     }
 
+    const requiredFields = [['modelId', 'modelName'], [], [], [], []];
```
petermuller, line 148: We can make streaming optional by setting it to false in the api model under the hood, that might simplify things, but I believe we also need modelType here too. as far as i can tell, required fields are modelId, modelName, streaming (which is easy to remove), and modelType (textgen/embedding)

lib/user-interface/react/src/shared/model/model-management.model.ts
```diff
@@ -167,8 +167,8 @@ export const containerConfigSchema = z.object({
 });
 
 export const ModelRequestSchema = z.object({
-    modelId: z.string().default(''),
-    modelName: z.string().default(''),
+    modelId: z.string().min(1).default(' '),
+    modelName: z.string().min(1).default(' '),
```
petermuller, line 171: Should these have default values at all? at minimum modelId should always be defined no matter the request. UpdateModel could optionally have modelName, but Id should still be there

estohlmann: This has to be done because of how we create the object for the form. I remove the ' ' when I create the form.

lib/user-interface/react/src/components/model-management/create-model/CreateModelModal.tsx
estohlmann: From a UI perspective there is no need to make this optional since it is a toggle. I can default it to false instead of true if you like

estohlmann: We also already default the picker to Textgen for model type.. So there isn't a way that it can be submitted without a type since it doesn't have a clear option

petermuller: Ah, sounds good- backend brain is not factoring in frontend reality :D

petermuller (approved):

PR #88, opened 2024-09-09
Pulling APIGW out from in front of ALB, adding admin auth via Secrets Manager

estohlmann:
Updating the ALB so that it isn't behind APIGW (doesn't work for model streaming) and storing internal admin token in Secrets Manager so that we can call model management apis from the state machine.


lambda/management_key.py
```diff
@@ -0,0 +1,51 @@
...
+    #         secret_binary = base64.b64decode(response['SecretBinary'])
+    #         return secret_binary.decode('utf-8')
+
+    # except ClientError as e:
+    #     # Handle ResourceNotFoundException specifically
+    #     if e.response['Error']['Code'] == 'ResourceNotFoundException':
+    #         print(f"Secret with name '{secret_name}' not found.")
+    #     else:
+    #         print(f"Unexpected error: {e}")
+    #     return None
```
petermuller, line 51: are we expecting this in a future revision?

lambda/models/state_machine/create_model.py
```diff
@@ -206,6 +218,9 @@ def handle_add_model_to_litellm(event: Dict[str, Any], context: Any) -> Dict[str
         model_name=event["modelId"],
         litellm_params=litellm_params,
     )
+
+    print(f"LITELLM RESP {litellm_response}")
```
petermuller: extra debugging line

lib/serve/index.ts
```diff
@@ -76,6 +84,57 @@ export class LisaServeApplicationStack extends Stack {
...
+        //     vpc: props.vpc.vpc,
+        // });
+
+        // managementKeySecret.grantRead(rotateManagementKeyLambda);
+
+        // new RotationSchedule(this, createCdkId([id, 'RotateManagementRotationSchedule']), {
+        //     secret: managementKeySecret,
+        //     rotationLambda: rotateManagementKeyLambda,
+        //     automaticallyAfter: Duration.days(1), // Rotate every 30 days
+        //   });
```
petermuller, line 123: for the next revision / later feature?

lib/serve/index.ts
```diff
@@ -76,6 +84,57 @@ export class LisaServeApplicationStack extends Stack {
...
+
+        // You can now use the `secret` variable to access the created secret
+        // For example, you might output the secret's ARN for reference
+        new CfnOutput(this, createCdkId([id, 'RotateManagementKey']), {
+            value: managementKeySecret.secretArn,
+            description: 'The ARN of the secret',
+        });
```
petermuller, line 130: is this necessary/used? or do we just use the reference in other parts of the cdk definition?

lib/serve/rest-api/src/api/endpoints/v2/litellm_passthrough.py
```diff
@@ -90,11 +96,42 @@ async def litellm_passthrough(request: Request, api_path: str) -> Response:
     litellm_path = f"{LITELLM_URL}/{api_path}"
     headers = dict(request.headers.items())
 
+    print(f"HEADERS OBJ {headers}")
+
+    logger.info("attempting passthrough")
+    print("attempting passthrough")
+    if not is_valid_management_token(headers):
+        logger.info("mgmt token not valid")
+        print("mgmt token not valid")
```
petermuller: additional logging things can be removed now

lib/serve/rest-api/src/api/endpoints/v2/litellm_passthrough.py
```diff
@@ -90,11 +96,42 @@ async def litellm_passthrough(request: Request, api_path: str) -> Response:
...
+            else:
+                raise HTTPException(status_code=HTTP_401_UNAUTHORIZED, detail="Not authenticated in passthrough - 127")
+
     # At this point in the request, we have already validated auth with IdP or persistent token. By using LiteLLM for
     # model management, LiteLLM requires an admin key, and that forces all requests to require a key as well. To avoid
     # soliciting yet another form of auth from the user, we add the existing LiteLLM key to the headers that go directly
     # to the LiteLLM instance.
     headers["Authorization"] = f"Bearer {LITELLM_KEY}"
+    logger.info(f"overriding authorization token with {LITELLM_KEY}")
+    print(f"overriding authorization token with {LITELLM_KEY}")
```
petermuller: at least one of these can be removed

lib/serve/rest-api/src/api/endpoints/v2/litellm_passthrough.py
```diff
@@ -106,3 +143,37 @@ async def litellm_passthrough(request: Request, api_path: str) -> Response:
...
+def is_valid_management_token(headers: dict[str, str]) -> bool:
+    """Return if API Token from request headers is valid if found."""
+    secret_tokens = refresh_management_tokens()
+    token = get_authorization_token(headers=headers, header_name="Authorization").strip()
+    logger.info(f"Checking if {token} is in {secret_tokens}")
+
+    print(f"Headers in is_valid_management_token {headers}")
+    print(f"Headers in is_valid_management_token {headers}")
+
+    print(f"Checking if {token} is in {secret_tokens}")
```
petermuller: we probably don't want to log the token here, right?

lib/serve/rest-api/src/api/routes.py
```diff
@@ -15,17 +15,29 @@
...
+    logger.info("Auth disabled")
+else:
+    security = OIDCHTTPBearer()
+    dependencies = [Depends(security)]
+    logger.info("Auth enabled")
+
+router.include_router(
+    litellm_passthrough.router, prefix="/v2/serve", tags=["litellm_passthrough"], dependencies=dependencies
+)
 router.include_router(litellm_passthrough.router, prefix="/v2/serve", tags=["litellm_passthrough"])
```
petermuller, line 41: duplicate router?

lambda/management_key.py
estohlmann: Yes I think so.. This was the auto refresh stuff Dustin was planning on doing. I can pull it out if you want.

lib/core/iam/ecs.json
```diff
@@ -191,6 +191,11 @@
           ]
         }
       }
+    },
+    {
+      "Effect": "Allow",
+      "Action": "secretsmanager:GetSecretValue",
+      "Resource": "*"
```
petermuller, line 198: this seems overly permissive. i realize this is a json file, but can we scope somehow?

lib/serve/rest-api/src/api/endpoints/v2/litellm_passthrough.py
estohlmann: Correct, I ripped all of this out.

lib/models/model-api.ts
```diff
@@ -190,11 +190,20 @@ export class ModelsApi extends Construct {
                                 lisaServeEndpointUrlPs.parameterArn
                             ],
                         }),
+                        new PolicyStatement({
+                            effect: Effect.ALLOW,
+                            actions: [
+                                'secretsmanager:GetSecretValue'
+                            ],
+                            resources: ['*'],
```
petermuller, line 198: since we know the secret, can we use that to get the arn? or craft the arn instead of a * policy?

lib/serve/index.ts
estohlmann: I don't think this is used

lambda/management_key.py
petermuller: if we'll enable later, we can keep for now

petermuller (dismissed):

lib/core/iam/ecs.json
estohlmann: Not sure how to fix this due to a circular dependency. Serve has a dependency on the IAM stack, any suggestions?

lib/models/model-api.ts
estohlmann: fixed

petermuller (approved):

PR #99, opened 2024-09-18
Adds new authentication method for model management API

dustins:
*Issue #, if available:*

*Description of changes:*
Starting with the 3.0 release, which introduced dynamic model management, we required admin authentication through an IdP to manage models. However, this approach doesn't work in certain scenarios. To address this, the PR introduces a new authentication method using the model management token secret.

*Testing:*
- I am able to create a model using the Model Management UI 
- I am able to create a model by submitting a POST request to the APIGW endpoint `/model` using the model management token
- I am not able create a model by submitting a POST request to the APIGW endpoint `/model` without the management token

### DIFF
--- a/lambda/authorizer/lambda_functions.py
+++ b/lambda/authorizer/lambda_functions.py
@@ -28,7 +28,7 @@ from utilities.common_functions import authorization_wrapper, get_id_token, retr
 
 logger = logging.getLogger(__name__)
 
-secrets_manager = boto3.client("secretsmanager", config=retry_config)
+secrets_manager = boto3.client("secretsmanager", region_name=os.environ["AWS_REGION"], config=retry_config)
 
 
 @authorization_wrapper
--- a/lambda/authorizer/lambda_functions.py
+++ b/lambda/authorizer/lambda_functions.py
@@ -18,12 +18,16 @@ import os
 import ssl
 from typing import Any, Dict
 
+import boto3
 import create_env_variables  # noqa: F401
 import jwt
 import requests
 from utilities.common_functions import authorization_wrapper, get_id_token
 
 logger = logging.getLogger(__name__)
+
+secrets_manager = boto3.client("secretsmanager", region_name=os.environ["AWS_REGION"])
+secret_tokens = []
 
 
 @authorization_wrapper
@@ -47,6 +51,11 @@ def lambda_handler(event: Dict[str, Any], context) -> Dict[str, Any]:  # type: i
     jwt_groups_property = os.environ.get("JWT_GROUPS_PROP", "")
 
     deny_policy = generate_policy(effect="Deny", resource=event["methodArn"])
+
+    if is_valid_management_token(id_token):
+        allow_policy = generate_policy(effect="Allow", resource=event["methodArn"], username="lisa-management-token")
+        logger.debug(f"Generated policy: {allow_policy}")
+        return allow_policy
 
     if jwt_data := id_token_is_valid(id_token=id_token, client_id=client_id, authority=authority):
         is_admin_user = is_admin(jwt_data, admin_group, jwt_groups_property)
@@ -134,3 +143,30 @@ def is_admin(jwt_data: dict[str, Any], admin_group: str, jwt_groups_property: st
         else:
             return False
     return admin_group in current_node
+
+
+def refresh_management_tokens() -> list[str]:
+    """Return secret management tokens if they exist."""
+    global secret_tokens
+
+    if len(secret_tokens) == 0:
+        try:
+            secret_tokens.append(
+                secrets_manager.get_secret_value(
+                    SecretId=os.environ.get("MANAGEMENT_KEY_NAME"), VersionStage="AWSCURRENT"
+                )["SecretString"]
+            )
+            secret_tokens.append(
+                secrets_manager.get_secret_value(
+                    SecretId=os.environ.get("MANAGEMENT_KEY_NAME"), VersionStage="AWSPREVIOUS"
+                )["SecretString"]
+            )
+        except Exception:
+            logger.info(f"No previous secret version for {os.environ.get('MANAGEMENT_KEY_NAME')}")
+
+    return secret_tokens
+
+
+def is_valid_management_token(id_token: str) -> bool:
+    """Return if API Token is valid."""
+    return id_token in refresh_management_tokens()
--- a/lambda/authorizer/lambda_functions.py
+++ b/lambda/authorizer/lambda_functions.py
@@ -16,18 +16,19 @@
 import logging
 import os
 import ssl
+from functools import cache
 from typing import Any, Dict
 
 import boto3
 import create_env_variables  # noqa: F401
 import jwt
 import requests
-from utilities.common_functions import authorization_wrapper, get_id_token
+from botocore.exceptions import ClientError
+from utilities.common_functions import authorization_wrapper, get_id_token, retry_config
 
 logger = logging.getLogger(__name__)
 
-secrets_manager = boto3.client("secretsmanager", region_name=os.environ["AWS_REGION"])
-secret_tokens = []
+secrets_manager = boto3.client("secretsmanager", config=retry_config)
 
 
 @authorization_wrapper
@@ -52,7 +53,7 @@ def lambda_handler(event: Dict[str, Any], context) -> Dict[str, Any]:  # type: i
 
     deny_policy = generate_policy(effect="Deny", resource=event["methodArn"])
 
-    if is_valid_management_token(id_token):
+    if id_token in get_management_tokens():
         allow_policy = generate_policy(effect="Allow", resource=event["methodArn"], username="lisa-management-token")
         logger.debug(f"Generated policy: {allow_policy}")
         return allow_policy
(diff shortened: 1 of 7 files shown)


test/cdk/stacks/core-api-base.test.ts
```diff
@@ -112,6 +112,6 @@ describe.each(regions)('API Core Nag Pack Tests | Region Test: %s', (awsRegion)
 
     test('NIST800.53r5 CDK NAG Errors', () => {
         const errors = Annotations.fromStack(stack).findError('*', Match.stringLikeRegexp('NIST.*'));
-        expect(errors.length).toBe(6);
+        expect(errors.length).toBe(7);
```
dustins, line 115: This change is a complaint that the authorizer is using an inline policy as a result of adding in `managementKeySecret.grantRead(authorizerLambda);`

lambda/authorizer/lambda_functions.py
```diff
@@ -134,3 +143,30 @@ def is_admin(jwt_data: dict[str, Any], admin_group: str, jwt_groups_property: st
         else:
             return False
     return admin_group in current_node
+
+
+def refresh_management_tokens() -> list[str]:
+    """Return secret management tokens if they exist."""
+    global secret_tokens
```
petermuller, line 150: lets rename to `get_management_tokens` and then remove the global variable since we don't have a "global" use case based on the code here

lambda/authorizer/lambda_functions.py
```diff
@@ -18,13 +18,17 @@
...
 
+import boto3
 import create_env_variables  # noqa: F401
 import jwt
 import requests
 from utilities.common_functions import authorization_wrapper, get_id_token
 
 logger = logging.getLogger(__name__)
 
+secrets_manager = boto3.client("secretsmanager", region_name=os.environ["AWS_REGION"])
```
petermuller, line 29: let's add the retry config in here too- we have a method for that, we just need to import it. examples:
https://github.com/awslabs/LISA/blob/develop/lambda/models/lambda_functions.py#L24
https://github.com/awslabs/LISA/blob/develop/lambda/models/lambda_functions.py#L64-L67

lambda/authorizer/lambda_functions.py
```diff
@@ -134,3 +143,30 @@ def is_admin(jwt_data: dict[str, Any], admin_group: str, jwt_groups_property: st
...
+                    SecretId=os.environ.get("MANAGEMENT_KEY_NAME"), VersionStage="AWSCURRENT"
+                )["SecretString"]
+            )
+            secret_tokens.append(
+                secrets_manager.get_secret_value(
+                    SecretId=os.environ.get("MANAGEMENT_KEY_NAME"), VersionStage="AWSPREVIOUS"
+                )["SecretString"]
+            )
+        except Exception:
+            logger.info(f"No previous secret version for {os.environ.get('MANAGEMENT_KEY_NAME')}")
```
petermuller, line 165: can we add the stack trace or actual error in the output too? if we get an exception for something that is not related to getting the previous key then we're outputting a possibly misleading error message.

lambda/authorizer/lambda_functions.py
```diff
@@ -134,3 +143,30 @@ def is_admin(jwt_data: dict[str, Any], admin_group: str, jwt_groups_property: st
...
+
+    return secret_tokens
+
+
+def is_valid_management_token(id_token: str) -> bool:
```
petermuller, line 170: from one of my other comments, we may be able to remove this

lib/api-base/authorizer.ts
```diff
@@ -82,12 +86,16 @@ export class CustomAuthorizer extends Construct {
...
                 JWT_GROUPS_PROP: config.authConfig!.jwtGroupsProperty,
+                MANAGEMENT_KEY_NAME: managementKeySecretNameStringParameter.stringValue
             },
             role: role,
             vpc: vpc,
             securityGroups: securityGroups,
         });
 
+        const managementKeySecret = Secret.fromSecretNameV2(this, createCdkId([id, 'managementKey']), managementKeySecretNameStringParameter.stringValue);
+        managementKeySecret.grantRead(authorizerLambda);
```
petermuller, line 97: is this correct? or should the read be granted to the authorizer lambda role? i normally would have expected that the execution role gets the permissions

lib/serve/index.ts
```diff
@@ -46,6 +46,8 @@ export class LisaServeApplicationStack extends Stack {
     public readonly restApi: FastApiContainer;
     public readonly modelsPs: StringParameter;
     public readonly endpointUrl: StringParameter;
+    public readonly managementKeySecret: Secret;
+    public readonly managementKeySecretNameStringParameter: StringParameter;
```
petermuller, line 50: why do we need to add these as externally readable members? i'd REALLY like to avoid exposing these, and instead just contrive the values in code, like you already did. I don't see any external references to these values (which is good), but because of the dangers of adding cross-stack dependencies that can block cdk updates (especially with constantly rotating values), I would much prefer to not expose these at all and keep them internal to this stack. (please see this change as a reference for my aversion to cross-stack param references: https://github.com/awslabs/LISA/commit/e2ae9efa5e169354f24f7d81ffac8007c6599d52 )

petermuller (changes requested):

lambda/utilities/common_functions.py
```diff
@@ -285,8 +285,12 @@ def get_id_token(event: dict) -> str:
...
         raise ValueError("Missing authorization token.")
 
-    token = auth_header[1]
-    return str(token)
+    if len(auth_header) == 1:
+        # secret management token won't be split into multiple segments
+        return str(auth_header[0])
+    else:
+        # Bearer tokens will have the token in the second segment
+        return str(auth_header[1])
```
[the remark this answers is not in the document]
dustins: I agree that is much better.

lib/api-base/authorizer.ts
dustins: It does add it to the lambda role. I can try adding it for the role directly though.

dustins: Actually since `role` is nullable and it works like this it is probably best to keep it as is since we know that the `authorizerLambda` will never be null.

lambda/authorizer/lambda_functions.py
```diff
@@ -134,3 +144,24 @@ def is_admin(jwt_data: dict[str, Any], admin_group: str, jwt_groups_property: st
...
     return admin_group in current_node
+
+
+@cache
+def get_management_tokens() -> list[str]:
+    """Return secret management tokens if they exist."""
+    secret_tokens: list[str] = []
+    secret_id = os.environ.get("MANAGEMENT_KEY_NAME")
+
+    if len(secret_tokens) == 0:
```
petermuller, line 155: this is guaranteed to be 0, so we can remove this check

lambda/utilities/common_functions.py
```diff
@@ -279,14 +280,14 @@ def get_id_token(event: dict) -> str:
...
     elif "Authorization" in event["headers"]:
-        auth_header = event["headers"]["Authorization"].split(" ")
+        auth_header = event["headers"]["Authorization"]
     else:
         raise ValueError("Missing authorization token.")
 
-    token = auth_header[1]
-    return str(token)
+    # remove bearer token prefix if present
+    return re.sub(r"Bearer ", "", str(auth_header), flags=re.I).strip()
```
petermuller, line 290: I'd prefer the removeprefix string method since it's a bit easier to read than this and doesn't import a library, but not a blocking comment. this still does the job- it's a stretch, but this may be a problem if somehow a user has a token that also contains "Bearer " in it (largely unlikely, but possible)

```
>>> import re
>>> re.sub(r"Bearer ", "", "Bearer Bearer 1", flags=re.I)
'1'
```

vs

```
>>> "Bearer Bearer 1".removeprefix("Bearer ")
'Bearer 1'
```

something to consider, i'll accept either way though- i'm sure secrets manager wouldn't give us that string specifically, and spaces probably shouldn't make it into an auth header definition

lib/api-base/authorizer.ts
petermuller: TIL. nice!

lambda/utilities/common_functions.py
dustins: I'll push the change with `removeprefix` but since there is no capitalization guarantees so it is actually between the two below.

```
>>> import re
>>> re.sub(r"^Bearer ", "", id_token, flags=re.I)
``` 

and

```
>>> id_token.removeprefix("Bearer ").removeprefix("bearer ")
```

lambda/authorizer/lambda_functions.py
```diff
@@ -16,15 +16,20 @@
...
 import create_env_variables  # noqa: F401
 import jwt
 import requests
-from utilities.common_functions import authorization_wrapper, get_id_token
+from botocore.exceptions import ClientError
+from utilities.common_functions import authorization_wrapper, get_id_token, retry_config
 
 logger = logging.getLogger(__name__)
 
+secrets_manager = boto3.client("secretsmanager", config=retry_config)
```
petermuller, line 31: region should still be here though, right?

petermuller: added it back in next commit.

petermuller (approved):

PR #114, opened 2024-09-25
ECS Config Refactor in config.yaml

estohlmann:
*Description of changes:*

When creating LISA 3.0 we removed the ability for users to define models and have them deployed at deploy time from the config file. Updating the schema to optionally provide the required fields to pull and validate model weights are present in the S3 bucket. Also redefined the schema structure to support the optional array and removed values. The new field will look like so:

```
  ecsModels:
    - modelName: meta-llama/Meta-Llama-3-8B
      baseImage: vllm/vllm-openai:v0.4.3
      inferenceContainer: vllm
    - modelName: mistralai/Mistral-7B-Instruct-v0.2
      baseImage: ghcr.io/huggingface/text-generation-inference:2.0.1
      inferenceContainer: tgi
    - modelName: e5-large-v2
      baseImage: ghcr.io/huggingface/text-generation-inference:2.0.1
      inferenceContainer: tei
    - modelName: google/flan-t5-xl
      baseImage: ghcr.io/huggingface/text-generation-inference:2.2.0
      inferenceContainer: tgi
```

Also prevented the roles for the models from automatically being created, as this will be done at deploy time through the model management UI/APIs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -842,7 +842,7 @@ you can do so.
       ```yaml
       ecsModels:
         - modelName: your-model-name
-          inferenceContainer: tgi 
+          inferenceContainer: tgi
           baseImage: ghcr.io/huggingface/text-generation-inference:2.0.1
       ```
 - If you are deploying the LISA Chat User Interface you can optionally specify the path to the pre-built
--- a/README.md
+++ b/README.md
@@ -372,15 +372,15 @@ restApiConfig:
 
 In the `ecsModels` section of `config.yaml`, allow our deployment process to pull the model weights for you.
 
-During the deployment proces, lisa will optionally attempt to download your model weights if you specify an optional`ecsModels`
-array, this will only work in non ADC regions. Specifically, see the `ecsModels` section of the [config.yaml](./config.yaml) file.
+During the deployment process, LISA will optionally attempt to download your model weights if you specify an optional `ecsModels`
+array, this will only work in non ADC regions. Specifically, see the `ecsModels` section of the [example_config.yaml](./example_config.yaml) file.
 Here we define the model name, inference container, and baseImage:
 
 ```yaml
 ecsModels:
   - modelName: your-model-name
     inferenceContainer: tgi
-    baseImage: ghcr.io/huggingface/text-generation-inference:1.0.2
+    baseImage: ghcr.io/huggingface/text-generation-inference:2.0.1
 ```
 
 ---
@@ -842,8 +842,8 @@ you can do so.
       ```yaml
       ecsModels:
         - modelName: your-model-name
-          inferenceContainer: tgi # vLLM-specific config
-          baseImage: ghcr.io/huggingface/text-generation-inference:1.0.2 # vLLM-specific config
+          inferenceContainer: tgi 
+          baseImage: ghcr.io/huggingface/text-generation-inference:2.0.1
       ```
 - If you are deploying the LISA Chat User Interface you can optionally specify the path to the pre-built
   website assets using the top level `webAppAssetsPath` parameter in `config.yaml`. Specifying this path
--- a/README.md
+++ b/README.md
@@ -370,34 +370,18 @@ restApiConfig:
 
 ## Step 9: Customize Model Deployment
 
-In the `ecsModels` section of `config.yaml`, configure the models you want to deploy.
+In the `ecsModels` section of `config.yaml`, allow our deployment process to pull the model weights for you.
 
-The configuration file will determine which models are deployed. In order to deploy an additional model or a
-different model the only required change is to the configuration file, as long as it is compatible with the inference
-container. Specifically, see the `ecsModels` section of the [config.yaml](./config.yaml) file.
-Here we define the model name, if we want to deploy, the type of instance we want to deploy to, the type of model
-(textgen or embedding), the inference container and then the containerConfig. There are many more parameters for the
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -370,35 +370,19 @@ restApiConfig:
...
-In the `ecsModels` section of `config.yaml`, configure the models you want to deploy.
+In the `ecsModels` section of `config.yaml`, allow our deployment process to pull the model weights for you.
 
-The configuration file will determine which models are deployed. In order to deploy an additional model or a
-different model the only required change is to the configuration file, as long as it is compatible with the inference
-container. Specifically, see the `ecsModels` section of the [config.yaml](./config.yaml) file.
-Here we define the model name, if we want to deploy, the type of instance we want to deploy to, the type of model
-(textgen or embedding), the inference container and then the containerConfig. There are many more parameters for the
-ecs models, many for autoscaling and health checks. However, let's focus on the model specific ones:
+During the deployment proces, lisa will optionally attempt to download your model weights if you specify an optional`ecsModels`
```
petermuller, line 375: s/lisa/LISA/g

README.md
```diff
@@ -857,17 +841,9 @@ you can do so.
...
-          streaming: true # remove option if modelType is 'embedding'
-          instanceType: g5.xlarge
-          inferenceContainer: vllm # vLLM-specific config
-          containerConfig:
-            image:
-              baseImage: vllm/vllm-openai:v0.5.0 # vLLM-specific config
-              path: lib/serve/ecs-model/vllm # vLLM-specific config
+        - modelName: your-model-name
+          inferenceContainer: tgi # vLLM-specific config
+          baseImage: ghcr.io/huggingface/text-generation-inference:1.0.2 # vLLM-specific config
```
petermuller, line 846: vLLM references on tgi options. also that container tag is way too old. just for examples, everything for TGI should be >=1.4.0

README.md
```diff
@@ -370,35 +370,19 @@ restApiConfig:
...
+array, this will only work in non ADC regions. Specifically, see the `ecsModels` section of the [config.yaml](./config.yaml) file.
+Here we define the model name, inference container, and baseImage:
 
 ```yaml
 ecsModels:
   - modelName: your-model-name
-    deploy: false
-    instanceType: g4dn.12xlarge
-    modelType: textgen
     inferenceContainer: tgi
```
petermuller, line 382: we can simplify even further by removing this portion from the config. we'd only need to check for the docker image since the tgi/tei/vllm options only matter for the path component at deployment time, but not for artifact collection time. all we would need are the model name (`$provider/$model` format) and the docker image/tag that we'd like to run with. even with the docker image- we don't do any validations on that initially that I'm aware of, and we just *assume* that the customer has access to the image at deployment time (during createmodel). so ultimately, this could just be a list of modelNames in huggingface $provider/$model format

README.md
```diff
@@ -370,35 +370,19 @@ restApiConfig:
...
-    containerConfig:
-      baseImage: ghcr.io/huggingface/text-generation-inference:1.0.2
-      environment:
-        QUANTIZE: bitsandbytes-nf4
-        MAX_CONCURRENT_REQUESTS: 128
-        MAX_INPUT_LENGTH: 1024
-        MAX_TOTAL_TOKENS: 2048
+    baseImage: ghcr.io/huggingface/text-generation-inference:1.0.2
```
petermuller, line 383: We have a few different version tags in the config file. can we standardize on a single tag version for the tgi and tei ones? (they won't be the same tags between tei/tgi, but all tgi tags should be the same for example)

README.md
```diff
@@ -370,35 +370,19 @@ restApiConfig:
...
+In the `ecsModels` section of `config.yaml`, allow our deployment process to pull the model weights for you.
 
-The configuration file will determine which models are deployed. In order to deploy an additional model or a
-different model the only required change is to the configuration file, as long as it is compatible with the inference
-container. Specifically, see the `ecsModels` section of the [config.yaml](./config.yaml) file.
-Here we define the model name, if we want to deploy, the type of instance we want to deploy to, the type of model
-(textgen or embedding), the inference container and then the containerConfig. There are many more parameters for the
-ecs models, many for autoscaling and health checks. However, let's focus on the model specific ones:
+During the deployment proces, lisa will optionally attempt to download your model weights if you specify an optional`ecsModels`
+array, this will only work in non ADC regions. Specifically, see the `ecsModels` section of the [config.yaml](./config.yaml) file.
```
petermuller, line 376: technically this file doesn't exist in the repo. let's just use backticks to display as `config.yaml` or something to indicate it's important, but let's not link directly to it, otherwise this is a 404 (which i realize is already in the readme... this is already a bug in the readme 😬 )

README.md
estohlmann: It was using the docker image and inference container in this section of the make file:

```
						echo "Converting and uploading safetensors for model: $(MODEL_ID)"; \
						tgiImage=$$(yq -r '[.$(ENV).ecsModels[] | select(.inferenceContainer == "tgi") | .baseImage] | first' $(PROJECT_DIR)/config.yaml); \
						echo $$tgiImage; \
						$(PROJECT_DIR)/scripts/convert-and-upload-model.sh -m $(MODEL_ID) -s $(MODEL_BUCKET) -a $$access_token -t $$tgiImage -d $$localModelDir; \
```

so rather than doing an entire refactor I was trying to keep the changes limited. I can see what the convert and upload script uses them for.

petermuller (approved):

Issue #92, opened 2024-09-12
Python 3.12 blocker in govcloud regions


david-saeger:
The Python 3.12 runtime is not currently available in GovCloud regions. Attempting to deploy on v3 the [xxx]-[xxx]-models-dev stack failed. The logical ID from CloudFormation of the error is `ModelsApidockerimagebuilderowlowlmodelsdevdockerimagebuilder295C9303` and the error message 
```
Resource handler returned message: "Access denied for runtime python3.12. Specify a different runtime and retry the request. (Service: Lambda, Status Code: 400, Request ID: 5b7c2626-7c1b-4d44-a48a-e7ae3ac403a6)" (RequestToken: b4dc9729-752e-1899-5caa-627ff04ede8b, HandlerErrorCode: InvalidRequest)
```

I see 3.12 was added as a dependency here https://github.com/awslabs/LISA/blob/0e824eb42dd779ab93c6976d2bcde94271c20f50/lib/models/docker-image-builder.ts#L115 and in a couple other places. Curious how deeply reliant v3 is on the latest python. Might have to fall back to v2 or come up with another plan.

petermuller:
Good catch! As far as I know, we're not reliant on anything in the most recent python version- I know just from personal projects that 3.12 breaks some things that I normally use (although I can't remember what those are at the moment 😬 )

Prior to this set of changes, we've been relying on the python version in the config.yaml file, which i've had success with 3.9 and 3.10, and I haven't tried 3.11. I think 3.9 or 3.10 would be safe versions to try instead of 3.12 if GovCloud is lacking the most recent versions.

if 3.9 or 3.10 work without a problem, then we should make the updates to the props to add the Config object and then we can plug in `config.lambdaConfig.pythonRuntime` instead of the hardcoded version.

Just based on the schema.ts file, the max we supported so far is 3.11, so this one was just an oversight.

---

tl;dr, I expect 3.9 or 3.10 to work fine instead of 3.12

petermuller:
Related review merged. We will now pull the Python version that is defined in config.yaml for the model deployment pieces. Closing this issue 👍

PR #118, opened 2024-09-27
Bug/prefix name clash

estohlmann:
*Issue #, if available:*

*Description of changes:*

Fixing an issue where adding role prefixes would cause LISA to not deploy. Also limiting the prefix fields to 20 characters.

### DIFF
--- a/lib/core/utils.ts
+++ b/lib/core/utils.ts
@@ -22,6 +22,8 @@ import * as cdk from 'aws-cdk-lib';
 import * as iam from 'aws-cdk-lib/aws-iam';
 
 import { Config } from '../schema';
+import { Effect, ManagedPolicy, PolicyDocument, PolicyStatement, Role, ServicePrincipal } from 'aws-cdk-lib/aws-iam';
+import { Construct } from 'constructs';
 
 const IAM_DIR = path.join(__dirname, 'iam');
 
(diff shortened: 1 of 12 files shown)


dustins (dismissed):

petermuller (dismissed):

petermuller:
Nice! We can close out https://github.com/awslabs/LISA/issues/92 with this set of changes too. Thanks for taking care of the python thing in here too

petermuller (dismissed):

lib/core/utils.ts
```diff
@@ -22,6 +22,8 @@ import * as cdk from 'aws-cdk-lib';
 import * as iam from 'aws-cdk-lib/aws-iam';
 
 import { Config } from '../schema';
+import { Effect, ManagedPolicy, PolicyDocument, PolicyStatement, Role, ServicePrincipal } from 'aws-cdk-lib/aws-iam';
```
petermuller, line 25: nitpick on duplicate iam imports- likely IDE added both- i'd try removing the star import instead of this one

lib/models/docker-image-builder.ts
```diff
@@ -77,6 +79,7 @@ export class DockerImageBuilder extends Construct {
         ec2InstanceProfileRole.attachInlinePolicy(ec2InstanceProfilePolicy);
 
         const role = new Role(this, createCdkId([stackName, 'docker_image_builder_role']), {
+            roleName: createCdkId([stackName, 'docker_image_builder_role']),
```
petermuller, line 82: not anything we need to address now, but as a backlog item or something later, we should standardize name formats. we have every example possible between capitals, dashes, underscores, and so on for logical and physical names

test/cdk/stacks/chat.test.ts
```diff
@@ -120,6 +120,6 @@ describe.each(regions)('Chat Nag Pack Tests | Region Test: %s', (awsRegion) => {
 
     test('NIST800.53r5 CDK NAG Errors', () => {
         const errors = Annotations.fromStack(stack).findError('*', Match.stringLikeRegexp('NIST.*'));
-        expect(errors.length).toBe(17);
+        expect(errors.length).toBe(14);
```
petermuller, line 123: Nice to see these numbers go down :D

petermuller (dismissed):
Some nitpicky comments in my latest review, but not anything that is blocking.

petermuller (approved):

PR #122, opened 2024-10-01
Bug: Fixed CDK stack dependency for secret management key

dustins:
*Issue #, if available:*

*Description of changes:*
With the latest changes to pull the secret management key into the APIGW Authorizer we now need to make sure that the Serve stack is created before API Base stack or else brand new deployments can fail. This change addresses that issue.

### DIFF
--- a/lib/serve/index.ts
+++ b/lib/serve/index.ts
@@ -17,7 +17,7 @@
 // LISA-serve Stack.
 import path from 'path';
 
-import { Stack, StackProps } from 'aws-cdk-lib';
+import { RemovalPolicy, Stack, StackProps } from 'aws-cdk-lib';
 import { AttributeType, BillingMode, Table, TableEncryption } from 'aws-cdk-lib/aws-dynamodb';
 import { Peer, Port, SecurityGroup } from 'aws-cdk-lib/aws-ec2';
 import { Credentials, DatabaseInstance, DatabaseInstanceEngine } from 'aws-cdk-lib/aws-rds';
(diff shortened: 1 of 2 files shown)


estohlmann (approved):

lib/serve/index.ts
```diff
@@ -91,6 +91,7 @@ export class LisaServeApplicationStack extends Stack {
                 excludePunctuation: true,
                 passwordLength: 16
             },
+            removalPolicy: RemovalPolicy.RETAIN_ON_UPDATE_OR_DELETE
```
petermuller, line 94: we're coming up with a new key every deployment so i'd expect that we're not ever "updating" it (secret name changes every time, so that's a [replacement](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-secretsmanager-secret.html#cfn-secretsmanager-secret-name) each time).

I think we can keep this for now, but for future iterations where we want the key to continuously update (if it's not actively doing that already), it's not super clear to me if we're going to have multiple key *names* to track  or if we will be updating the version/secret within a specific secret name.

Not really anything to action on right now, but something we should add for a backlog item to track to ensure that the key is working the way we expect it to.

petermuller (changes requested):
It looked like there was an added stack dependency before the rebase. Is that intentional? Do we want to *add* the dependency or remove it?

petermuller (approved):
Ignore my last comment, looks like the rebase brought in the change that was already there. Validated in the tree that the dependency change exists.

lib/serve/index.ts
dustins: This was leftover from testing having the secret be named the same every time. It was an attempt to workaround the issue where the Authorizer environment variables weren't staying up-to-date with the latest secret name. With the addition of the dependency though I believe this is solved now though and this can be removed.

PR #130, opened 2024-10-05
Add permissionsBoundaryAspect compatibility to model stacks

petermuller:
This set of changes allows the custom IaC Aspect to affect the stacks we create through the model management API as well, which up until this point we've missed supporting there. The CDK deployer function originally bundled all of the CDK node packages as part of the function, and adding one more dependency caused us to exceed the 250MB limit for functions. This was not something we could work around, so the function was converted to a Docker Image Function so that we had more wiggle room on it (limit is now 10GB).

The only change required to make the Models stacks include the custom Aspect was to version bump the IAC package because a recent change included a uniqueness string to alleviate the issues we were seeing with deeply nested Constructs. A later change will be needed to propagate this into the main deployment codebase as well.

Additionally, added payload logging to assist in debugging the deployer docker lambda function.

### DIFF
--- a/ecs_model_deployer/src/index.ts
+++ b/ecs_model_deployer/src/index.ts
@@ -51,6 +51,7 @@ const createWritableEnv = () => {
 };
 
 export const handler = async (event: any) => {
+    console.log(`Event payload: ${JSON.stringify(event)}`);
     if (!event.modelConfig) {
         console.log(`modelConfig not provided in ${JSON.stringify(event)}`);
         throw new Error('modelConfig not provided');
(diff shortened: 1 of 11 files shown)


petermuller:
Converted back to draft. Looking at the CDK labs package, it seems that if we update the version number, we should get a fix that adds a uniqueness disambiguator to the role names. Going to try reverting my changes to the models stack while strictly updating the version number for the cdklabs package.

ecs_model_deployer/src/index.ts
```diff
@@ -51,6 +51,7 @@ const createWritableEnv = () => {
 };
 
 export const handler = async (event: any) => {
+    console.log(`Event payload: ${JSON.stringify(event)}`);
```
estohlmann, line 54: Do we still want this

PR #142, opened 2024-10-09
Remove "path" from API model

petermuller:
The "path" component of the ContainerConfig object was redundant and unnecessarily complicated for customers to use. Historically, it referred to the path location of Dockerfiles from the root of the repo to where the model definitions were, but since we add customization on top of containers, we end up requiring that the paths are used. We specify in another option whether the model is using TGI/TEI/vLLM, and for all of these, there's a 1:1 match for it and its corresponding "path" option. The container spinup will fail if these don't match.

As part of the 3.0 release, the path changed even further to be relative to a folder inside of the root, which made it difficult to document and describe how to use it. This change removes it altogether and instead requires the backend to know where the files are, simplifying the request for customers, and preventing them from making more mistakes when creating a new LISA-hosted model.

### DIFF
--- a/README.md
+++ b/README.md
@@ -593,7 +593,7 @@ curl -s -H "Authorization: Bearer <admin_token>" -X GET https://<apigw_endpoint>
         }
       },
       "containerConfig": {
-        "baseImage": {
+        "image": {
           "baseImage": "vllm/vllm-openai:v0.5.0",
           "path": "vllm",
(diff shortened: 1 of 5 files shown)


petermuller (commented):
I can't review a PR that I created, but Evan your changes look fine to me 👍

estohlmann (approved):

PR #151, opened 2024-10-17
Unified error handling for Model Create/Update form

dustins:
*Issue #, if available:*

*Description of changes:*
The backend returns with different error responses depending on the type of error. Most errors are JSON responses (`{"message": string}`) but validation errors are slightly more complicated so multiple validation errors can be represented...

```json
{
  "detail": [
    {
      "loc": [
        "body",
        "size"
      ],
      "msg": "value is not a valid integer",
      "type": "type_error.integer"
    }
  ],
  "type": "RequestValidationError"
}
```

This PR updates the Model Create/Update form to handle the two types of responses possible.

### DIFF
--- a/lambda/models/lambda_functions.py
+++ b/lambda/models/lambda_functions.py
@@ -19,6 +19,8 @@ from typing import Annotated, Union
 import boto3
 import botocore.session
 from fastapi import FastAPI, Path, Request
+from fastapi.encoders import jsonable_encoder
+from fastapi.exceptions import RequestValidationError
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import JSONResponse
 from mangum import Mangum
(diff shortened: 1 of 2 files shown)


lambda/models/lambda_functions.py
```diff
@@ -63,6 +65,14 @@ async def model_not_found_handler(request: Request, exc: ModelNotFoundError) ->
     return JSONResponse(status_code=404, content={"message": str(exc)})
 
 
+@app.exception_handler(RequestValidationError)  # type: ignore
+async def validation_exception_handler(request: Request, exc: RequestValidationError):
+    """Handle exception when request fails validation and and translate to a 422 error."""
+    return JSONResponse(
```
estohlmann, line 71: Do we know that the only exceptions that will be caught by this are 422 errors? There aren't in lingering issues that could throw runtime exceptions (500 errors) that could be caught by this and accidentally translated to 422s?

dustins: By default FastAPI [already returns](https://github.com/fastapi/fastapi/blob/master/fastapi/exception_handlers.py#L20-L26) a [422](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422) for `RequestValidationError`…which seems appropriate. I wouldn't expect anything that is supposed to be a 5xx to be throwing a `RequestValidationError`. This is really re-implementing the default functionality for clarity by having all our exception handling in one place. I'd hope to start adding "type" to all the error responses so it is easier to discriminate between them in the future easily on the front end.

estohlmann (approved):

PR #152, opened 2024-10-18
Dynamic block device volume size and entry point changes

estohlmann:
A request was made to support block device volume size being configurable, added optional API param to accomplish this. 

For the TGI entrypoint:
- Added number of shards and CUDA settings (increase model memory footprint by distributing model across available GPU's)
- Fixed quantization not being applied by text-generation-launcher (variable being set but not applied by text-generation-launcher)


bedanley (dismissed):

ecs_model_deployer/src/lib/schema.ts
```diff
@@ -417,6 +417,7 @@ const MetricConfigSchema = z.object({
 * @property {MetricConfig} metricConfig - Metric configuration for auto scaling.
 */
 const AutoScalingConfigSchema = z.object({
+    blockDeviceVolumeSize: z.number().min(1).default(30),
```
dustins, line 420: I know there isn't a "good" solution to this but min 1 seems like a bad idea to allow doesn't it? Considering this field can't be updated later I'd think it would be better to keep the min as the default (30) and only allow larger values. Just a thought, but not blocking if you disagree so I'm still approving.

dustins (dismissed):

ecs_model_deployer/src/lib/schema.ts
estohlmann: Agreed, thanks for calling this out. I found that the value has to be >= 30 when doing some testing.. Need to update

dustins (approved):

PR #154, opened 2024-10-22
Enabling headless deployment in make file

estohlmann:
Enabling headless deployment in makefile. You can do this by running `make deploy HEADLESS=true`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -256,7 +256,7 @@ define print_config
 endef
 
 ## Deploy all infrastructure
-deploy:
+deploy: dockerCheck dockerLogin cleanMisc modelCheck buildEcsDeployer
 	$(call print_config)
 ifneq (,$(findstring true, $(HEADLESS)))
 	npx cdk deploy ${STACK} $(if $(PROFILE),--profile ${PROFILE}) --require-approval never -c ${ENV}='$(shell echo '${${ENV}}')';
--- a/Makefile
+++ b/Makefile
@@ -240,96 +240,43 @@ listStacks:
 buildEcsDeployer:
 	@cd ./ecs_model_deployer && npm install && npm run build
 
+define print_config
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -241,7 +242,13 @@ buildEcsDeployer:
 
 ## Deploy all infrastructure
 deploy: dockerCheck dockerLogin cleanMisc modelCheck buildEcsDeployer
-ifdef PROFILE
+ifneq (,$(findstring true, $(HEADLESS)))
+	if [ -z ${PROFILE} ]; then \
```
dustins, line 246: Is this indentation actually off or just how github is showing it?

Makefile
```diff
@@ -241,7 +242,13 @@ buildEcsDeployer:
...
+        npx cdk deploy ${STACK} -c ${ENV}='$(shell echo '${${ENV}}')'; \
+    else \
+        npx cdk deploy ${STACK} --profile ${PROFILE} -c ${ENV}='$(shell echo '${${ENV}}')'; \
```
bedanley, line 249: We will want `cdk deploy --require-approval never` added so the deployer doesn't have to confirm stack changes.

Makefile
```diff
@@ -241,7 +242,13 @@ buildEcsDeployer:
...
+    fi;
+else ifdef PROFILE
 	@printf "\n \
```
bedanley, line 252: We should still print out the app Infrastructure properties regardless of headless mode

Makefile
estohlmann: makefile syntax can be quirky
- if you want a line of code to be interpreted as make code it must only be indented with spaces.
- if you want a line of code to be interpreted as bash code it must only be indented with tabs

bedanley (approved):

PR #155, opened 2024-10-23
Feature/private vpc and subnets

estohlmann:
Enabling users to optionally set subnets during deployment

### DIFF
--- a/ecs_model_deployer/src/lib/ecs-model.ts
+++ b/ecs_model_deployer/src/lib/ecs-model.ts
@@ -15,7 +15,7 @@
 */
 
 // ECS Model Construct.
-import { ISecurityGroup, IVpc } from 'aws-cdk-lib/aws-ec2';
+import { ISecurityGroup, IVpc, SubnetSelection } from 'aws-cdk-lib/aws-ec2';
 import { AmiHardwareType } from 'aws-cdk-lib/aws-ecs';
 import { Bucket } from 'aws-cdk-lib/aws-s3';
 import { Construct } from 'constructs';
@@ -40,6 +40,7 @@ type ECSModelProps = {
     modelConfig: ModelConfig;
     securityGroup: ISecurityGroup;
     vpc: IVpc;
+    subnetSelection?: SubnetSelection | undefined;
 } & BaseProps;
 
 /**
@@ -56,7 +57,7 @@ export class EcsModel extends Construct {
    */
     constructor (scope: Construct, id: string, props: ECSModelProps) {
         super(scope, id);
-        const { config, modelConfig, securityGroup, vpc } = props;
+        const { config, modelConfig, securityGroup, vpc, subnetSelection } = props;
 
         const modelCluster = new ECSCluster(scope, `${id}-ECC`, {
             config,
@@ -74,6 +75,7 @@ export class EcsModel extends Construct {
             },
             securityGroup,
             vpc,
+            subnetSelection
         });
 
         // Single bucket for all models
--- a/ecs_model_deployer/src/lib/ecs-model.ts
+++ b/ecs_model_deployer/src/lib/ecs-model.ts
@@ -40,7 +40,7 @@ type ECSModelProps = {
     modelConfig: ModelConfig;
     securityGroup: ISecurityGroup;
     vpc: IVpc;
-    subnetSelection?: SubnetSelection | undefined;
+    subnetSelection?: SubnetSelection;
 } & BaseProps;
 
 /**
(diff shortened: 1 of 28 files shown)


ecs_model_deployer/src/lib/ecs-model.ts
```diff
@@ -40,6 +40,7 @@ type ECSModelProps = {
     modelConfig: ModelConfig;
     securityGroup: ISecurityGroup;
     vpc: IVpc;
+    subnetSelection?: SubnetSelection | undefined;
```
dustins, line 43: I don't think `| undefined` is necessary when defined with `?` right?

ecs_model_deployer/src/lib/lisa_model_stack.ts
```diff
@@ -38,13 +38,28 @@ export class LisaModelStack extends Stack {
...
+        if (props.config.subnetIds && props.config.subnetIds.length > 0) {
+            const iSubnets: Map<number, ISubnet> = new Map();
+            props.config.subnetIds.forEach((subnet, index) => {
+                const importedSubnet = Subnet.fromSubnetId(this, index.toString(), subnet);
+                iSubnets.set(index, importedSubnet);
+            });
+
+            subnetSelection = {
+                subnets: Array.from(iSubnets!.values()),
+            };
```
dustins, line 52: This could be needed for a reason I don't understand but couldn't this be simplified by something like below? Seems like going to a map is an unnecessary step. 

```
subnetSelection = {
  subnets: props.config.subnetIds.map((subnet, index) => Subnet.fromSubnetId(this, index.toString(), subnet))
};

File: lib/models/ecs-model-deployer.ts [hunk 3]
```diff
@@ -69,10 +71,12 @@ export class ECSModelDeployer extends Construct {
             memorySize: 1024,
             role: role,
             environment: {
-                'LISA_VPC_ID': props.vpcId,
+                'LISA_VPC_ID': props.vpc.vpc.vpcId,
```
dustins, line 74: Everywhere else we're doing `props.vpc?.blah`, should this not be `props.vpc?.vpc` as well if it is potentially undefined?

lib/api-base/authorizer.ts
```diff
@@ -33,11 +34,13 @@ import { Secret } from 'aws-cdk-lib/aws-secretsmanager';
...
  * @property {IRole} role - Execution role for lambdas
  * @property {ISecurityGroup[]} securityGroups - Security groups for Lambdas
+ * @property {Map<number, ISubnet>} importedSubnets for Lambdas
  */
 type AuthorizerProps = {
     role?: IRole;
-    vpc?: IVpc;
+    vpc?: Vpc;
     securityGroups?: ISecurityGroup[];
+    importedSubnets?: Map<number, ISubnet>;
```
dustins, line 43: Is this being used indirectly somewhere? I could have missed it but I don't see it being used in this commit.

ecs_model_deployer/src/lib/schema.ts
```diff
@@ -618,6 +618,7 @@ const RawConfigSchema = z
                 instanceProfilePrefix: z.string().optional(),
             })
             .optional(),
+        subnetIds: z.array(z.string()).optional(),
```
bedanley, line 621: I would make this more explicit: `subnetIds: z.array(z.string().startsWith('subnet-')).optional()`

bedanley: Thinking about this some more, we should force `vpcId` to be required if this exists. We can do this with https://github.com/colinhacks/zod#refine as we've done with other conditional properties.

lib/networking/vpc/index.ts
```diff
@@ -54,9 +62,30 @@ export class Vpc extends Construct {
 
         let vpc: IVpc;
         if (config.vpcId) {
+            /** Imports VPC for use by application if supplied, else creates a VPC. */
```
bedanley, line 65: Nit: Use multiline coments instead of jsdocs within a function

lib/networking/vpc/index.ts
```diff
@@ -44,6 +46,12 @@ export class Vpc extends Construct {
     /** Security groups for application. */
     public readonly securityGroups: SecurityGroups;
 
+    /** Created from deployment configured Subnets for application. */
+    public readonly subnetGroup?: SubnetGroup | undefined;
+
+    /** Imported Subnets for application. */
+    public readonly subnetSelection?: SubnetSelection | undefined;
```
bedanley, line 53: No need for undefined with optional members

bedanley (approved):

dustins (approved):

PR #157, opened 2024-10-24
Config schema and CDK updates

estohlmann:
Updated schema, expected config parsing, and CDK deployment to account for these changes


lib/serve/index.ts
```diff
@@ -174,8 +172,8 @@ export class LisaServeApplicationStack extends Stack {
                 username: username,
                 passwordSecretId: litellmDbPasswordSecret.secretName,
                 dbHost: litellmDb.dbInstanceEndpointAddress,
-                dbName: rdsConfig.dbName,
-                dbPort: rdsConfig.dbPort,
+                dbName: 'postgres',
+                dbPort: 5432,
```
dustins, line 176: Why are we making this *less* configurable?

PR #169, opened 2024-11-01
Feature/in app config management

estohlmann:
Added ability for administrators to configure system banner settings and which components of the chat UI are enabled for all users.


lib/user-interface/react/src/shared/modal/confirmation-modal.tsx
```diff
@@ -27,7 +27,7 @@ export type ConfirmationModalProps = {
     resourceName: string;
     onConfirm: () =>  MutationActionCreatorResult<any>;
     postConfirm?: CallbackFunction;
-    description?: string;
+    description?: any;
```
bedanley, line 30: What would description be if not a `string`?

test/cdk/stacks/chat.test.ts
```diff
@@ -105,12 +105,12 @@ describe.each(regions)('Chat Nag Pack Tests | Region Test: %s', (awsRegion) => {
     //TODO Update expect values to remediate CDK NAG findings and remove debug
     test('AwsSolutions CDK NAG Warnings', () => {
         const warnings = Annotations.fromStack(stack).findWarning('*', Match.stringLikeRegexp('AwsSolutions-.*'));
-        expect(warnings.length).toBe(1);
+        expect(warnings.length).toBe(2);
```
bedanley, line 108: Take a quick pass to see if we can knock out any of these nag findings with the newly created table.

bedanley (approved):

lib/user-interface/react/src/shared/modal/confirmation-modal.tsx
estohlmann: I am passing HTML for the diff in config

Updated to:
`description?: string | ReactElement;`

dustins (approved):

bedanley (approved):

PR #172, opened 2024-11-05
Feature/role overrides

bedanley:
Adding role overrides for LISA. Role overrides allow a user to provide their own roles in the event they don't want LISA to automatically create them. This pulls in existing roles (where possible), instead of creating them via CDK. 


dustins (approved):

lib/docs/admin/api.md
```diff
@@ -0,0 +1,364 @@
+
```
estohlmann, line 1: Looks like you had a bad rebase for the docs directory. All the changes in here should already exist on develop in different file names.

bedanley: Rebased off of latest and removed docs

estohlmann (approved):

PR #196, opened 2024-11-25
Feature/sg overrides

bedanley:
Add Security Group overrides to all stacks.

This allows users to provide their own Security Groups, instead of using LISA created SGs. 


example_config.yaml
```diff
@@ -13,6 +13,13 @@ s3BucketModels: hf-models-gaiic
 # subnets:
 #  - subnetId:
 #    ipv4CidrBlock:
+# securityGroupConfig: The below values are none/some/all
```
estohlmann, line 16: based on our convo this should be none or all

lib/stages.ts
```diff
@@ -265,6 +310,16 @@ export class LisaServeApplicationStage extends Stage {
             });
         }
 
+        if (config.securityGroupConfig) {
+            stacks.forEach((lisaStack) => {
+                if (config.securityGroupConfig?.modelSgId) {
```
estohlmann, line 315: Based on modelSgId being required if securityGroupConfig is defined, I think this check isn't needed

dustins (approved):

lib/schema.ts
```diff
@@ -45,16 +46,43 @@ export enum EcsSourceType {
...
+ * @property {string} openSearchSgId - Security Group ID.
+ * @property {string} pgVectorSgId - Security Group ID.
+ */
+export const SecurityGroupConfigSchema = z.object({
+    modelSgId: z.string().startsWith('sg-'),
+    restAlbSgId: z.string().startsWith('sg-'),
+    lambdaSgId: z.string().startsWith('sg-'),
+    liteLlmDbSgId: z.string().startsWith('sg-'),
+    openSearchSgId: z.string().startsWith('sg-').optional(),
+    pgVectorSgId: z.string().startsWith('sg-').optional(),
```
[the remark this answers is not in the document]
estohlmann: It's never too late ;)

estohlmann (approved):

PR #202, opened 2024-12-03
Closing file context manager if upload is success

estohlmann:
Close the file context manager if file ingestion/upload is successful

### DIFF
--- a/lib/user-interface/react/src/components/chatbot/Chat.tsx
+++ b/lib/user-interface/react/src/components/chatbot/Chat.tsx
@@ -434,7 +434,7 @@ export default function Chat ({ sessionId }) {
                         new LisaChatMessage({
                             type: 'ai',
                             content: result,
-                            metadata: metadata,
+                            metadata: useRag ? { ...metadata, ...prev.history[prev.history.length - 1].metadata } : metadata,
                         }),
                     ),
                 }));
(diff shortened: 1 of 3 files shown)


lib/user-interface/react/src/components/chatbot/Chat.tsx
```diff
@@ -434,7 +434,7 @@ export default function Chat ({ sessionId }) {
                         new LisaChatMessage({
                             type: 'ai',
                             content: result,
-                            metadata: metadata,
+                            metadata: useRag ? { ...metadata, ...prev.history[prev.history.length - 1].metadata } : metadata,
```
bedanley, line 437: Are we handling an empty history array?

estohlmann: It won't be empty is RAG is used because when we get the context back it sets a chat history message already

dustins: …if we're already talking about this line it might be cleaner to use `prev.history.slice(-1)[0].metadata`? I already approved so not blocking at all…more of a preference.

estohlmann: Honestly looking at this code today, it all needs to be re-written, it is pretty nasty. Let's make note of this and I will get a task written to do some cleanup around this chat piece.

dustins (approved):

bedanley (approved):

PR #209, opened 2024-12-06
Feature/partition make

bedanley:
Add partition and domain overrides when deploying LISA. With these additions, we can now run `make bootstrap` and `make deploy` in regions not in commercial AWS with the overrides `PARTITION` and `DOMAIN`. 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,11 +56,20 @@ ifeq (${REGION},)
 $(error region must be set in command line using REGION variable or config files)
 endif
 
-# URL_SUFFIX - used for the docker login
+ifeq (${PARTITION},)
+PARTITION := $(shell cat $(PROJECT_DIR)/config-custom.yaml | yq .partition )
+endif
+ifeq (${PARTITION}, null)
+PARTITION := aws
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -56,11 +56,20 @@ ifeq (${REGION},)
 $(error region must be set in command line using REGION variable or config files)
 endif
 
-# URL_SUFFIX - used for the docker login
+ifeq (${PARTITION},)
```
estohlmann, line 59: Is there not a way to get the partition or domain from CDK like we do in `core/utils.ts`

```
.replace(/\${AWS::Partition}/gi, cdk.Aws.PARTITION)
```

estohlmann: This is clearly using the ts library but wasn't sure if we could use the command line tool to figure this out since we already require customers to supply the region

bedanley: I can retrieve the partition using AWS cli within the makefile, but it does add that as a new dependency (which presumably is already installed).

estohlmann (approved):

PR #215, opened 2024-12-18
Feature/rag list api

bedanley:
Adding RAG List/Delete APIs

The addition of the following APIs will allow a user to list documents for a given repository and collection, as well as delete a select document in the collection. Deleting a document will delete all chunked parts of the original document from the collection. 

### DIFF
--- a/lambda/repository/lambda_functions.py
+++ b/lambda/repository/lambda_functions.py
@@ -292,7 +292,6 @@ def delete_document(event: dict, context: dict) -> Dict[str, Any]:
 
     query_string_params = event["queryStringParameters"]
     collection_id = query_string_params["collectionId"]
-    repository_type = query_string_params["repositoryType"]
     document_id = query_string_params.get("documentId")
     document_name = query_string_params.get("documentName")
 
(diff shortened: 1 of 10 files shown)


lambda/repository/lambda_functions.py
```diff
@@ -325,3 +445,44 @@ def presigned_url(event: dict, context: dict) -> dict:
...
+    collection_id = query_string_params.get("collectionId")
+    docs: list[RagDocument] = doc_repo.list_all(repository_id, collection_id)
+
+    # return docs
+
+    id_token = get_id_token(event)
+    embeddings = _get_embeddings(model_name=collection_id, id_token=id_token)
+    vs = get_vector_store_client(repository_id, index=collection_id, embeddings=embeddings)
+
+    doc_search = vs.similarity_search(
```
estohlmann, line 481: Do we really want to do a similarity search here to list out just the documents that have been loaded?

bedanley: This should have been removed. Good catch.

estohlmann (approved):

PR #220, opened 2025-01-10
Feature/rest api sdk

bedanley:
Created a number of SDK clients, which include the majority of API Gateway endpoints:
- RAG listDocs, deleteDoc
- Model listModels, listEmbeddedModels
- Config listConfigs
- Session listSessions / history
- Repos listRepos

In addition, a small set of simple integration tests have been added to verify API interactions. 

The size of most lambdas have been reduced by removing unnecessary python libraries that are already included with a Python lambda deployment.  


lambda/utilities/common_functions.py
```diff
@@ -360,7 +360,7 @@ def get_session_id(event: dict) -> str:
 
 def get_groups(event: Any) -> List[str]:
     """Get user groups from event."""
-    groups: List[str] = event.get("requestContext", {}).get("authorizer", {}).get("groups", [])
+    groups: List[str] = json.loads(event.get("requestContext", {}).get("authorizer", {}).get("groups", "[]"))
```
estohlmann, line 363: Do we want the [] inclosed in " here

dustins (approved):

lib/core/layers/fastapi/requirements.txt
```diff
@@ -1,4 +1,4 @@
-boto3==1.34.131
+# boto3==1.34.131 // Provided by Lambda
```
estohlmann, line 1: Let's just remove these then, same with above.

estohlmann (approved):

PR #223, opened 2025-01-15
Bug/Store RAG Document Metadata Subdocs in Separate Table 

bedanley:
Saving RAG SubDocument IDs in separate table to overcome Dynamo record limit. RAG Document records are now stored in 2 tables: 1 for metadata and 1 for subdocument chunks. This should allow for much larger documents to be managed and support quicker queries to retrieve RAG document metadata. 


lib/rag/index.ts
```diff
@@ -109,15 +111,54 @@ export class LisaRagStack extends Stack {
...
             },
+            sortKey: {
+                name: 'sk',
+                type: AttributeType.STRING
+            },
+            billingMode: dynamodb.BillingMode.PAY_PER_REQUEST,
+            encryption: dynamodb.TableEncryption.AWS_MANAGED,
+            removalPolicy: config.removalPolicy,
         });
+        // const docTable = new Table(this, ragTableName, {
```
estohlmann, line 138: If this isn't needed let's remove

lambda/repository/rag_document_repo.py
```diff
@@ -175,18 +175,69 @@ def list_all(self, repository_id: str, collection_id: str) -> List[RagDocument]:
...
+            response = self.doc_table.query(
                 KeyConditionExpression=Key("pk").eq(pk),
-                ExclusiveStartKey=response["LastEvaluatedKey"],
             )
-            docs.extend(response["Items"])
+            docs: list[RagDocumentDict] = response.get("Items", [])
 
-        return docs
+            # Handle paginated Dynamo results
+            while "LastEvaluatedKey" in response:
```
estohlmann, line 186: This is fine for now, but wondering if we will want to keep the request paginated so that we don't crash a UI with an extensive amount of files in the store.

estohlmann (dismissed):

lib/rag/index.ts
bedanley: Removed

lambda/models/domain_objects.py
```diff
@@ -312,6 +315,22 @@ class IngestionType(Enum):
     MANUAL = "manual"
 
 
+RagDocumentDict: TypeAlias = Dict[str, Any]
```
dustins, line 318: TIL…I didn't know you could do this in python!

lambda/repository/rag_document_repo.py
dustins: Yeah I think this should be paginated instead of just loading everything. We can just provide the pagination key to the frontend and we'll be a glorified proxy. I think it is pretty standard to not have a `list_all` that returns an unlimited number of resources at once.

estohlmann (approved):

lambda/repository/rag_document_repo.py
dustins: I agree, we should just be proxying dynamo and using their continuation tokens.

lambda/models/domain_objects.py
```diff
@@ -322,16 +341,48 @@ class RagDocument(BaseModel):
     document_name: str
     source: str
     username: str
-    sub_docs: List[str] = Field(default_factory=lambda: [])
+    subdocs: List[str] = Field(default_factory=lambda: [], exclude=True)
```
dustins, line 344: IMHO this feels kinda clunky having this on this object considering it isn't persisted or returned to the frontend. I'd just use the result of `join_docs` result directly instead of trying to store it on the RagDocument. That way you can stop worrying about that in the `find*` methods too.

lambda/models/domain_objects.py
```diff
@@ -322,16 +341,48 @@ class RagDocument(BaseModel):
...
         return f"{repository_id}#{collection_id}"
+
+    def chunk_doc(self, chunk_size: int = 1000) -> list[RagSubDocument]:
+        """Chunk the document into smaller sub-documents."""
+        chunked_docs: list[RagSubDocument] = []
+        for i in range(0, len(self.subdocs), chunk_size):
+            subdocs = self.subdocs[i : i + chunk_size]
+            chunk = RagSubDocument(document_id=self.document_id, subdocs=subdocs, index=i)
+            chunked_docs.append(chunk)
+        return chunked_docs
```
dustins, line 368: Not that this will be too slow but because this list can easily be in the thousands I'd make it a generator. Something like this…

```
def chunk_doc(self, chunk_size: int = 1000):
    """Chunk the document into smaller sub-documents, yielding each chunk."""
    for i in range(0, len(self.subdocs), chunk_size):
        subdocs = self.subdocs[i : i + chunk_size]
        chunk = RagSubDocument(document_id=self.document_id, subdocs=subdocs, index=i)
        yield chunk  # Yielding each chunk lazily
```

lambda/repository/rag_document_repo.py
bedanley: Replaced with paginated results

lambda/models/domain_objects.py
```diff
@@ -322,16 +341,48 @@ class RagDocument(BaseModel):
     document_name: str
     source: str
     username: str
-    sub_docs: List[str] = Field(default_factory=lambda: [])
+    subdocs: List[str] = Field(default_factory=lambda: [], exclude=True)
+    chunk_size: int
+    chunk_overlap: int
```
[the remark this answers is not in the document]
bedanley: I'm looking into storing a map for the strategy. It may not make it into this release.

lambda/models/domain_objects.py
bedanley: Keeping for now since this is a convenient way to pass along the subdocs as part of the document. It isn't queried with the API unless explicitly requested.

lambda/models/domain_objects.py
bedanley: Neat. Done.

lambda/models/domain_objects.py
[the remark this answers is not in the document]
bedanley: Updated to store map

estohlmann (approved):

lambda/repository/pipeline_ingest_documents.py
```diff
@@ -159,8 +159,7 @@ def handle_pipeline_ingest_documents(event: Dict[str, Any], context: Any) -> Dic
             document_name=key,
             source=docs[0][0].metadata.get("source"),
             subdocs=all_ids,
-            chunk_size=chunk_size,
-            chunk_overlap=chunk_overlap,
+            chunk_strategy={"chunk_size": str(chunk_size), "chunk_overlap": str(chunk_overlap)},
```
dustins, line 162: Only because this is kinda committing us to these values. I think we should have "name" for the type of strategy so we can use it as a discriminator. Also just a small thing but we might want to stop prefixing everything with `chunk_` too since we're already in a chunk strategy object.

bedanley: Removed prefix and added StrategyType to entries.

estohlmann (approved):

PR #225, opened 2025-01-16
Update new session link to use base url

bedanley:
Update new session link to use base url


lib/user-interface/react/src/components/chatbot/Sessions.tsx
```diff
@@ -118,7 +118,8 @@ export function Sessions () {
                     {
                         id: 'title',
                         header: 'Title',
-                        cell: (e) => <Link variant='primary' href={`#/chatbot/${e.sessionId}`}>{e.history[0].content || 'No Content'}</Link>,
+                        cell: (e) => <Link variant='primary'
+                            href={`${window.env.API_BASE_URL}#/chatbot/${e.sessionId}`}>{e.history[0].content || 'No Content'}</Link>,
```
dustins, line 122: We do this wrong all over I'm sure, but we should probably use [getBaseURI()](https://github.com/awslabs/LISA/blame/develop/lib/user-interface/react/src/components/utils.ts#L29)

bedanley: Updated in all locations

dustins (approved):

PR #231, opened 2025-01-29
Create RAG deletion pipeline

bedanley:
Create RAG deletion pipeline based on repository configuration. 

This adds the ability to remove S3 documents when removed from RAG vector stores. 
 

lib/user-interface/react/src/components/chatbot/FileUploadModals.tsx
```diff
@@ -114,7 +122,8 @@ export function ContextUploadModal ({
                     <SpaceBetween direction='horizontal' size='xs'>
                         <Button
                             onClick={async () => {
-                                const successfulUploads = await handleUpload(selectedFiles, handleError, processFile, [FileTypes.TEXT], 204800);
+                                const files = selectedFiles.map((f) => renameFile(f));
```
dustins, line 125: Hard to tell from just this but are we giving feedback that they clicked the button? Like temporarily disabling it while this uploads?

lib/schema.ts
```diff
@@ -539,22 +539,27 @@ const OpenSearchExistingClusterConfig = z.object({
     endpoint: z.string(),
 });
 
+export const PipelineConfigSchema = z.object({
+    chunkOverlap: z.number(),
```
dustins, line 543: I think we're moving to storing the strategy instead of just assuming chunking. Should we update this to match?

lib/user-interface/react/src/components/chatbot/FileUploadModals.tsx
bedanley: This is renaming the file being uploaded with a prepended timestamp to prevent collisions with other rag uploaded files. 

And yes, the upload button is disabled until they select a file with `disabled={selectedFiles.length === 0}`. After all files are uploaded, the modal closes.

lib/schema.ts
bedanley: We can update when that change is made. This is simply exposing the current PipelineConfig as a ts type.

dustins (approved):

PR #236, opened 2025-01-31
Add document link downloader to Document Library

bedanley:
Add document link downloader to Document Library. Create a new presigned url for downloading a single RAG document from S3. 


lib/user-interface/react/src/components/document-library/DocumentLibraryComponent.tsx
```diff
@@ -84,13 +88,19 @@ export function DocumentLibraryComponent ({ repositoryId }: DocumentLibraryCompo
             selection: { trackBy: 'document_id' },
         },
     );
-
+    const [isDownloading, setIsDownloading] = useState<boolean>(false);
+    const [getDownloadUrl] = useLazyDownloadRagDocumentQuery();
```
dustins, line 92: I think instead of worrying about `isDownloading` you can use `isFetching` from the hook. I *think* it would look something like this…

```typescript
    const [isDownloading, setIsDownloading] = useState<boolean>(false);
    const [getDownloadUrl, {isFetching: isDownloading}] = useLazyDownloadRagDocumentQuery();
```

dustins (dismissed):

lib/user-interface/react/src/components/document-library/DocumentLibraryComponent.tsx
bedanley: Agreed. Updated.

dustins (approved):

PR #240, opened 2025-02-12
Vector Store Management Page

bedanley:
Create new UI for managing Vector Stores through configuration page


lib/user-interface/react/src/components/configuration/RepositoryActions.tsx
```diff
@@ -0,0 +1,154 @@
...
+            disabled: selectedItems.length !== 1,
+        }];
+
+    return (
+        <ButtonDropdown
+            items={items}
+            variant='primary'
+            disabled={!selectedRepo}
+            loading={isDeleteLoading || isUpdating}
+            onItemClick={(e) =>
```
dustins, line 117: Can we type this so we're not using `e: any` below?

dustins (approved):

PR #245, opened 2025-02-15
Legacy Version Sets

dustins:
*Issue #, if available:*

*Description of changes:*
As part of our work on dynamic vector stores, we removed support for YAML-configured vector stores. However, to ensure users retain access to their existing vector stores, this update restores that functionality while preventing the creation of new ones. Users can still update and remove their existing stores.


lib/rag/index.ts
```diff
@@ -139,12 +151,6 @@ export class LisaRagStack extends Stack {
             RAG_DOCUMENT_TABLE: docMetaTable.tableName,
             RAG_SUB_DOCUMENT_TABLE: subDocTable.tableName,
             ADMIN_GROUP: config.authConfig!.adminGroup,
-            REGISTERED_REPOSITORIES_PS_NAME: `${config.deploymentPrefix}/registeredRepositories`,
-            MANAGEMENT_KEY_SECRET_NAME_PS: `${config.deploymentPrefix}/managementKeySecretName`,
-            RDS_CONNECTION_INFO_PS_NAME: `${config.deploymentPrefix}/${connectionParamName}`,
-            OPENSEARCH_ENDPOINT_PS_NAME: `${config.deploymentPrefix}/lisaServeRagRepositoryEndpoint`,
-            REGISTERED_REPOSITORIES_PS_PREFIX: `${config.deploymentPrefix}/LisaServeRagConnectionInfo/`,
-            LOG_LEVEL: config.logLevel,
```
bedanley, line 147: I would keep these all here so they can be used for all the RAG stack components.

lib/rag/index.ts
```diff
@@ -302,4 +320,293 @@ export class LisaRagStack extends Stack {
...
+            }
+
+            // ensure the entry gets updated in the database
+            new customResources.AwsCustomResource(this, createCdkId(['InsertRAGConfig', ragConfig.repositoryId]), {
+                onCreate: {
+                    service: 'DynamoDB',
+                    action: 'putItem',
+                    parameters: {
+                        TableName: ragRepositoryConfigTable.tableName,
+                        Item: {
```
bedanley, line 482: Why not pass the entire item to `toDynamoDBItem`

lib/rag/state_machine/legacy-ingest-pipeline.ts
```diff
@@ -0,0 +1,340 @@
...
+type PipelineConfig = {
+    chunkOverlap: number;
+    chunkSize: number;
+    embeddingModel: string;
+    s3Bucket: string;
+    s3Prefix: string;
+    trigger: string;
+};
+
+type RdsConfig = {
```
bedanley, line 53: You can import`RdsConfig` and `PipelineConfig` from configSchema

bedanley (approved):

lib/rag/state_machine/legacy-ingest-pipeline.ts
dustins: Thanks! I think this was just leftover from how it was in 3.5

lib/rag/index.ts
dustins: I think that wouldn't match the existing items in the database.

lib/rag/index.ts
dustins: Nothing is using the first two at least but the last two can stay I think.

bedanley (approved):

PR #247, opened 2025-02-16
Add ability to delete legacy vector stores through UI

dustins:
*Issue #, if available:*

*Description of changes:*
Although legacy vector stores cannot be edited through the UI, we aim to enable their deletion. While this will not remove the resources, it will render them inactive and allow for their destruction during the next deployment. This PR also includes the following UI improvements:

- Disabled the edit option for legacy repositories.
- Updated the create repository form to include descriptions.
- Added a mandatory acknowledgment when deleting legacy repositories, informing users that a deployment is required to actually remove AWS resources.

Screenshot 2025-02-16 at 21 43 00

Screenshot 2025-02-16 at 21 11 54

### DIFF
--- a/lambda/repository/lambda_functions.py
+++ b/lambda/repository/lambda_functions.py
@@ -247,9 +247,7 @@ def list_all(event: dict, context: dict) -> List[Dict[str, Any]]:
     registered_repositories = vs_repo.get_registered_repositories()
     admin_override = is_admin(event)
     return [
-        repo
-        for repo in registered_repositories
-        if admin_override or user_has_group(user_groups, repo["allowedGroups"])
+        repo for repo in registered_repositories if admin_override or user_has_group(user_groups, repo["allowedGroups"])
     ]
 
 
--- a/lambda/repository/lambda_functions.py
+++ b/lambda/repository/lambda_functions.py
@@ -706,7 +706,7 @@ def delete(event: dict, context: dict) -> Any:
         return {"status": "success", "executionArn": response["executionArn"]}
 
 
-def _remove_legacy(repository_id: str):
+def _remove_legacy(repository_id: str) -> None:
     registered_repositories = ssm_client.get_parameter(Name=os.environ["REGISTERED_REPOSITORIES_PS"])
     registered_repositories = json.loads(registered_repositories["Parameter"]["Value"])
     updated_repositories = [repo for repo in registered_repositories if repo.get("repositoryId") != repository_id]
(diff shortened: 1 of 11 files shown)


lambda/repository/lambda_functions.py
```diff
@@ -245,7 +245,11 @@ def list_all(event: dict, context: dict) -> List[Dict[str, Any]]:
     """
     user_groups = get_groups(event)
     registered_repositories = vs_repo.get_registered_repositories()
-    return [repo for repo in registered_repositories if user_has_group(user_groups, repo["allowedGroups"])]
+    return [
+        repo
+        for repo in registered_repositories
+        if is_admin(event) or user_has_group(user_groups, repo["allowedGroups"])
```
bedanley, line 251: is_admin does parse the current lambda context. I would reuse the output for each repo in case we have to iterate over many repos.

lambda/repository/lambda_functions.py
```diff
@@ -683,16 +687,35 @@ def delete(event: dict, context: dict) -> Any:
         raise ValidationError("repositoryId is required")
 
     repository = vs_repo.find_repository_by_id(repository_id=repository_id, raw_config=True)
+    if repository.get("editable", False) is False:
```
bedanley, line 690: Is the `editable` field something that needs to be stored with all new repos? Could we invert that so that we only need it with legacy repos?

lib/user-interface/react/src/shared/form/array-input.tsx
```diff
@@ -15,18 +15,19 @@
...
 import React, { ReactElement } from 'react';
-import FormField from '@cloudscape-design/components/form-field';
+import FormField, { FormFieldProps } from '@cloudscape-design/components/form-field';
 import Input from '@cloudscape-design/components/input';
 import { Button, SpaceBetween } from '@cloudscape-design/components';
 
-type ArrayInputProps = {
-    label: string;
+type ArrayInputProps = FormFieldProps & {
+    // label: string;
```
bedanley, line 23: remove

bedanley (approved):

lambda/repository/lambda_functions.py
dustins: I could see changing it to something like "legacy"

lib/user-interface/react/src/shared/form/array-input.tsx
dustins: Weird I removed that already, maybe I forgot to commit it!

lambda/repository/lambda_functions.py
bedanley: If we could omit the value for new repos as default and only add it to legacy as a flag to the config entry, that would be preferred.

dustins: I just pushed some changes for this. Old repositories have legacy set. The "legacy" property only is applied to legacy repositories so it'll never make its way back into the config because of an update. This is better for sure.

PR #249, opened 2025-02-17
Disable deprecated TLS versions 1.0/1.1

bedanley:
Disable deprecated TLS versions 1.0/1.1

### DIFF
--- a/lib/api-base/ecsCluster.ts
+++ b/lib/api-base/ecsCluster.ts
@@ -38,7 +38,11 @@ import {
     Protocol,
     Volume,
 } from 'aws-cdk-lib/aws-ecs';
-import { ApplicationLoadBalancer, BaseApplicationListenerProps } from 'aws-cdk-lib/aws-elasticloadbalancingv2';
+import {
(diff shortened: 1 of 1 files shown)


lib/api-base/ecsCluster.ts
```diff
@@ -289,6 +293,7 @@ export class ECSCluster extends Construct {
             certificates: ecsConfig.loadBalancerConfig.sslCertIamArn
                 ? [{ certificateArn: ecsConfig.loadBalancerConfig.sslCertIamArn }]
                 : undefined,
+            sslPolicy: ecsConfig.loadBalancerConfig.sslCertIamArn ? SslPolicy.RECOMMENDED_TLS : SslPolicy.RECOMMENDED,
```
dustins, line 296: What is the difference between these two settings?

bedanley: Recommeded allows http, while recommenedTLS is only latest TLS.

dustins (approved):

PR #250, opened 2025-02-17
Bug/fixes 20240217

dustins:
*Issue #, if available:*

*Description of changes:*
Misc bug fixes…

- Parameter for sdk layer already existed so it couldn't be moved to a different stack without renaming.
- `allowedGroups` is/was optional so usage needs to default to an empty array
- handle existing pipelines that didn't yet define the `autoRemove` attribute
- fixed calling wrong action when removing rag repository from dynamodb


lib/core/index.ts
```diff
@@ -120,7 +120,7 @@ export class CoreStack extends cdk.Stack {
         });
 
         new StringParameter(this, createCdkId([config.deploymentName, config.deploymentStage, 'SdkLayer']), {
-            parameterName: `${config.deploymentPrefix}/layerVersion/sdk`,
+            parameterName: `${config.deploymentPrefix}/layerVersion/sdk-2`,
```
bedanley, line 123: Nit: Rename `lisa-sdk` or `pythonSdk`

bedanley (approved):

PR #259, opened 2025-02-21
Feature/refactor schema

bedanley:
Refactor Schema to share across all modules


lib/api-base/fastApiContainer.ts
```diff
@@ -108,7 +106,7 @@ export class FastApiContainer extends Construct {
                     cooldown: 60,
                     defaultInstanceWarmup: 60,
                     metricConfig: {
-                        AlbMetricName: 'RequestCountPerTarget',
+                        albMetricName: 'RequestCountPerTarget',
```
estohlmann, line 109: I don't see this change added to the FE, we should make sure that this works still for the API call.

estohlmann (approved):

lib/api-base/fastApiContainer.ts
bedanley: This change was added to the FE schema: https://github.com/awslabs/LISA/pull/259/files#diff-447aa0570f55cd991d82bbe6650a50b6ae2dc8e75a965c4b54d76dafa7678e3fR389

PR #261, opened 2025-03-03
Properly display error messages from proxied calls

estohlmann:
LISA has not properly displayed error messages for quite some time, this is because all responses were returning http_status 200. Updated BE to pass back the status code from the proxied call.

Screenshot 2025-03-03 at 2 28 17 PM

### DIFF
--- a/lib/user-interface/react/src/components/chatbot/Chat.tsx
+++ b/lib/user-interface/react/src/components/chatbot/Chat.tsx
@@ -271,7 +271,11 @@ export default function Chat ({ sessionId }) {
                 await memory.saveContext({input: params.inputs.input}, {output: resp.join('')});
                 setIsStreaming(false);
             } catch (exception) {
-                notificationService.generateNotification('An error occurred while processing your request.', 'error');
+                setSession((prev) => ({
+                    ...prev,
+                    history: prev.history.slice(0, -1),
+                }));
+                notificationService.generateNotification('An error occurred while processing your request.', 'error', undefined, exception.error?.message ? <p>{JSON.stringify(exception.error.message)}</p> : undefined);
             }
         };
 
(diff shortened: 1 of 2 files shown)


lib/user-interface/react/src/components/chatbot/Chat.tsx
```diff
@@ -271,7 +271,11 @@ export default function Chat ({ sessionId }) {
                 await memory.saveContext({input: params.inputs.input}, {output: resp.join('')});
                 setIsStreaming(false);
             } catch (exception) {
-                notificationService.generateNotification('An error occurred while processing your request.', 'error');
+                setSession((prev) => ({
+                    ...prev,
+                    history: prev.history.slice(0, -1),
+                }));
+                notificationService.generateNotification('An error occurred while processing your request.', 'error', undefined, exception.error?.message ? <p>{JSON.stringify(exception.error.message)}</p> : undefined);
```
bedanley, line 278: Based on the attached image, it seems like we can get an error like `{message: { message: "ERROR"} }`. We could look into recursively finding the message error if it exists. It is fine as is since we don't control the exception from external services like Bedrock.

bedanley (approved):

lib/user-interface/react/src/components/chatbot/Chat.tsx
```diff
@@ -290,7 +294,7 @@ export default function Chat ({ sessionId }) {
                     ),
                 }));
             } catch (exception) {
-                notificationService.generateNotification('An error occurred while processing your request.', 'error');
+                notificationService.generateNotification('An error occurred while processing your request.', 'error', undefined, exception.error?.message ? <p>{JSON.stringify(exception.error.message)}</p> : undefined);
```
dustins, line 297: Just confirming, `exception.error.message` is really a JSON object and not just a string already?

dustins (approved):

lib/user-interface/react/src/components/chatbot/Chat.tsx
estohlmann: Some errors it is a string and for others it is a JSON object, depend on the model provider and where the error is being thrown.

I was getting black screened depending on which error was thrown. Bedrock exceptions return a JSON object, LiteLLM errors are strings, until we support custom parsing this seems to be the best path forward to at least display something useful.

PR #277, opened 2025-03-27
Prompt Templates

dustins:
*Issue #, if available:*

*Description of changes:*
This update introduces a new Prompt Library UI area in LISA Chat, accessible via a new link at the top of the interface. The library allows users to manage their private prompts, share them globally or within specific groups, and copy prompts directly for use in conversations. Admins can enable or disable the library in Configuration. Users can create, edit, and delete their own prompts, and admins have the ability to modify all prompts. Additionally, users may view and select prompts directly at the start of conversations.

This also includes some minor UI improvements such as a new landing page, a more concise menu bar, and removal of unnecessary breadcrumbs, and renaming of chatbot to AI Assistant.

**New landing screen (with updated menu bar)**

**Prompt template editor**

**Prompt template library**

**Prompt template create/update**
| Private | Public | Selectively shared |
|--|--|--|
|  |  |  |

**Admin configuration**


lib/user-interface/react/src/App.tsx
```diff
@@ -101,80 +101,90 @@ function App () {
...
-                                    <Configuration setNav={setNav} />
-                                </AdminRoute>
-                            }
-                        />
-                    </Routes>
-                }
-            />
-            {confirmationModal && <ConfirmationModal {...confirmationModal} />}
-            {config?.configuration.systemBanner.isEnabled && <SystemBanner position='BOTTOM' />}
+            <ConfigurationContext.Provider value={config}>
```
dustins, line 104: I updated this to create a provider for our whole app. Instead of each component that needs the config loading it themselves.

lambda/prompt_templates/lambda_functions.py
```diff
@@ -0,0 +1,221 @@
...
+        return item
+
+    raise ValueError(f"Not authorized to get {prompt_template_id}.")
+
+
+def is_member(user_groups: List[str], prompt_groups: List[str]) -> bool:
+    if "lisa:public" in prompt_groups:
+        return True
+
+    return set(user_groups) & set(prompt_groups)
```
estohlmann, line 113: Strange that this doesn't return a bool

dustins: Good callout. I wasn't thinking and _felt_ this would be coerced into a bool. I updated this to do the coercion explicitly.

lambda/prompt_templates/lambda_functions.py
```diff
@@ -0,0 +1,221 @@
...
+    return model.model_dump()
+
+
+@api_wrapper
+def update(event: dict, context: dict) -> Any:
+    """Update an existing prompt template in DynamoDB."""
+    user_id = get_username(event)
+    prompt_template_id = get_prompt_template_id(event)
+    body = json.loads(event["body"], parse_float=Decimal)
+    model = PromptTemplateModel(**body)
```
estohlmann, line 155: Are we using model here because this was grabbed from that part of the baseline or because it's a model object? Seems to be a confusing name

lib/user-interface/react/src/components/chatbot/Chat.tsx
```diff
@@ -71,18 +77,17 @@ export default function Chat ({ sessionId }) {
             isFetching: state.isFetching,
             data: (state.data || []).filter((model) => model.modelType === ModelType.textgen && model.status === ModelStatus.InService),
         })});
-    const [getConfiguration] = useLazyGetConfigurationQuery();
-    const [config, setConfig] = useState<IConfiguration>();
     const [chatConfiguration, setChatConfiguration] = useState<IChatConfiguration>(baseConfig);
```
estohlmann, line 80: Is this really needed?

lib/user-interface/react/src/components/chatbot/Chat.tsx
```diff
@@ -71,18 +77,17 @@ export default function Chat ({ sessionId }) {
...
 
     const [userPrompt, setUserPrompt] = useState('');
     const [fileContext, setFileContext] = useState('');
 
     const [sessionConfigurationModalVisible, setSessionConfigurationModalVisible] = useState(false);
-    const [promptTemplateEditorVisible, setPromptTemplateEditorVisible] = useState(false);
+    const [promptTemplateKey, setPromptTemplateKey] = useState(new Date().toISOString());
```
estohlmann, line 86: Is this really needed?

lambda/prompt_templates/lambda_functions.py
dustins: Because it is a model object (extends pydantic's BaseModel). If you think that is confusing I'll change it. It felt verbose to say `prompt_template_model` but I'll conceed it may be too generically named especially in a project that overloads the term model.

lib/user-interface/react/src/components/chatbot/Chat.tsx
dustins: I'd argue yes. I'm using that so that we can easily reset the prompt template editor (as changing the key causes the component to be unmounted and a new instance used). This mounts a new instance of the component that doesn't have any pre-existing local state (like partially filled in text fields)

estohlmann: Got it. This is a different pattern than we use for our other modals so was curious the thought here. Thanks!

estohlmann (approved):

PR #279, opened 2025-03-28
UI Cleanup/Standardization

dustins:
*Issue #, if available:*

*Description of changes:*
This PR includes several small updates to enhance UI consistency and usability:  

- **Remove timestamp prefix from RAG document names**  
  - The timestamp prefix is unnecessary and does not match what users uploaded, making it confusing.  
  - This change removes the prefix in the UI, but we should also consider saving documents in DynamoDB without it.  
- **Improve resource listing behavior**  
  - Automatically refresh resource listings if they are in a non-terminal state.  
- **Standardize button labels**  
  - Ensure consistency in using either “New” or “Create” when adding new resources.  
- **Improve modal exit terminology**  
  - Replace “Abandon” with “Discard changes” for better clarity and consistency with common UI standards.  
- **Adjust action menu layout**  
  - Refresh buttons should always be positioned as the leftmost item in action menus.  
- **Enhance model object safety**  
  - Implement safe traversal of model objects, ensuring that features arrays are handled gracefully when not present.  
- **Refine resource creation buttons**  
  - Remove the “+” symbol from resource creation buttons to align with standard UI practices and reduce unnecessary space usage.  
- **Standardize public resource display**  
  - Use `(public)` as the consistent notation for publicly accessible resources.  
- **Make DockerImageBuilder volume size configurable**  
  - Allow configuration of the volume size (defaulting to the original value), as it is short-lived but may need adjustments for different use cases.  

These refinements improve clarity, consistency, and usability across the application.


Makefile
```diff
@@ -199,7 +199,7 @@ modelCheck:
 					then \
 						mkdir -p $$localModelDir; \
 						echo "What is your huggingface access token? "; \
-						read -s access_token; \
+						read access_token; \
```
dustins, line 202: `-s` isn't available on every platform

estohlmann (approved):

PR #283, opened 2025-04-01
Track session settings

estohlmann:
Storing session chat configuration so if user goes back to session, that data doesn't need to be updated again.

Updated session display to pull info out of table and created actions menu to contain items like delete/download session.

Also pruned down list session payload to not include all chat history.

Screenshot 2025-04-01 at 11 00 45 AM


batzela (approved):
Looks good to me.

bedanley (dismissed):

bedanley (approved):

PR #284, opened 2025-04-01
Revert "Bump vite from 6.2.2 to 6.2.4 in /lib/user-interface/react"

estohlmann:
Reverts awslabs/LISA#282


batzela (commented):
Looks good to me.

PR #291, opened 2025-04-10
Unit tests for Repos and Prompt Templates

estohlmann:
*Description of changes:*

Unit test coverage at ~26%

### DIFF
--- a/test/lambda/test_prompt_templates_lambda.py
+++ b/test/lambda/test_prompt_templates_lambda.py
@@ -1,0 +1,623 @@
+#   Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+#   Licensed under the Apache License, Version 2.0 (the "License").
+#   You may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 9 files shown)


test/lambda/test_prompt_templates_lambda.py
```diff
@@ -0,0 +1,689 @@
...
+        "title": "Test Template",
+        "revision": 1,
+        "latest": True,
+        "type": "persona",
+        "body": "Test prompt template body",
+    }
+
+
+@pytest.fixture(scope="function")
+def aws_credentials():
```
batzela, line 155: Why do these need to be set here when they are set at the top of the file?

test/lambda/test_repository_lambda.py
```diff
@@ -0,0 +1,1069 @@
...
+    else:
+        return MagicMock()  # Return a generic MagicMock for other services
+
+
+# Ensure mock_boto3_client is used for all boto3.client calls
+patch("boto3.client", side_effect=mock_boto3_client).start()
+
+
+@pytest.fixture(scope="function")
+def aws_credentials():
```
batzela, line 306: Same comment as above

batzela (dismissed):

batzela (approved):
Looks good to me!

PR #297, opened 2025-04-16
Feature/Isolated region deployments

bedanley:
- Update pathing for lambda assets to be absolute based on current location
- Allow overriding deployment asset locations to provide preconstructed assets (lambda, docker, s3 bucket, etc) 
- Update packaging of npm module to include built artifacts


lib/models/docker-image-builder.ts
```diff
@@ -53,9 +54,9 @@ export class DockerImageBuilder extends Construct {
         const { config } = props;
 
         const ec2DockerBucket = new Bucket(this, createCdkId([stackName, 'docker-image-builder-ec2-bucket']));
-
+        const ecsModelPath = ECS_MODEL_PATH;
```
estohlmann, line 57: I don't think we need to use this var if we aren't doing anything with it

bedanley: Leaving for now. We will want to add a configuration override like other pathing.

PR #298, opened 2025-04-17
Refactor stack to use constructs

bedanley:
- Refactor Stacks to use Constructs for creating stack resources

The majority of changes are moving the resources constructed in a Stack to a new Construct object. 


lib/user-interface/userInterfaceConstruct.ts
```diff
@@ -0,0 +1,277 @@
...
+        });
+    }
+}
+
+/**
+ * Copy files recursively from source directory to destination directory.
+ * @param {string} sourceDir - Source directory to copy from.
+ * @param {string} targetDir - Target directory to copy to.
+ */
+function copyDirRecursive (sourceDir: string, targetDir: string): void {
```
drduhe, line 259: Here, is it possible to just use cp directly rather than have a custom function for copying directories?

```
import { cp } from 'fs/promises';

/**
 * Recursively copies a directory.
 */
await cp('sourceDir', 'targetDir', { recursive: true });
```

lib/docs/docConstruct.ts
```diff
@@ -0,0 +1,201 @@
...
+import { CfnOutput, RemovalPolicy, Stack, StackProps } from 'aws-cdk-lib';
+import { AwsIntegration, EndpointType, RestApi } from 'aws-cdk-lib/aws-apigateway';
+import { IRole, Role, ServicePrincipal } from 'aws-cdk-lib/aws-iam';
+import { BlockPublicAccess, Bucket, BucketEncryption, IBucket } from 'aws-cdk-lib/aws-s3';
+import { BucketDeployment, Source } from 'aws-cdk-lib/aws-s3-deployment';
+import { Construct } from 'constructs';
+import { BaseProps } from '../schema';
+import { Roles } from '../core/iam/roles';
+
+const HERE: string = path.resolve(__dirname);
```
drduhe, line 29: I feel like we should be setting global paths based on the ROOT_DIR of the entire project rather than reading them in relatively from each file - but I am not sure you need to address that in this PR. But something like whats below could make it much cleaner everywhere:

``paths.ts``
```
 import path from 'path';
import fs from 'fs';

/**
 * ROOT_DIR is the base of your project. You could optionally override this with an env var.
 */
export const ROOT_DIR = path.resolve(__dirname);

/**
 * Utility: Resolve a path from the project root.
 */
export function fromRoot(...segments: string[]): string {
  return path.resolve(ROOT_DIR, ...segments);
}

/**
 * Optional: Check if a path exists (can be useful for conditional builds or logging).
 */
export function pathExists(...segments: string[]): boolean {
  return fs.existsSync(fromRoot(...segments));
}

/**
 * Directory Shortcuts
 */
export const LAYERS_DIR = fromRoot('layers');
export const SDK_DIR = fromRoot('..', 'lisa-sdk');

/**
 * Specific Layer Paths
 */
export const COMMON_LAYER_PATH = path.join(LAYERS_DIR, 'common');
export const FASTAPI_LAYER_PATH = path.join(LAYERS_DIR, 'fastapi');
export const AUTHORIZER_LAYER_PATH = path.join(LAYERS_DIR, 'authorizer');

/**
 * Export as grouped object for optional destructuring
 */
export const Paths = {
  ROOT: ROOT_DIR,
  LAYERS: LAYERS_DIR,
  SDK: SDK_DIR,
  COMMON_LAYER: COMMON_LAYER_PATH,
  FASTAPI_LAYER: FASTAPI_LAYER_PATH,
  AUTHORIZER_LAYER: AUTHORIZER_LAYER_PATH,
};
```

``usage``
```
import { COMMON_LAYER_PATH, pathExists, fromRoot } from '../paths';

console.log('Common Layer Path:', COMMON_LAYER_PATH);

if (!pathExists('layers', 'common')) {
  throw new Error('Missing required common layer directory!');
}

const buildScriptPath = fromRoot('scripts', 'build.ts');
```

lib/core/apiBaseConstruct.ts
```diff
@@ -0,0 +1,84 @@
...
+import { Role } from 'aws-cdk-lib/aws-iam';
+import { ITable } from 'aws-cdk-lib/aws-dynamodb';
+
+export type LisaApiBaseProps = {
+    vpc: Vpc;
+    tokenTable: ITable | undefined;
+} & BaseProps &
+    StackProps;
+
+export class LisaApiBaseConstruct extends Construct {
```
drduhe, line 33: Looks like we are missing TSDoc comments on this class.

lib/core/apiDeploymentConstruct.ts
```diff
@@ -0,0 +1,58 @@
...
+
+import { BaseProps } from '../schema';
+import { StringParameter } from 'aws-cdk-lib/aws-ssm';
+
+export type LisaApiDeploymentProps = {
+    restApiId: string;
+} & BaseProps &
+  StackProps;
+
+export class LisaApiDeploymentConstruct extends Construct {
```
drduhe, line 29: Same thing, TSDoc comments.

drduhe (commented):
LGTM but made some comments on how some general things could be improved. I would wait till others more experienced on the code base give their thumbs up. I am still getting familiarized with the logic everywhere.

lib/docs/docConstruct.ts
bedanley: I like consolidating the paths into a single file. I added this because the "root" path is lost if this is pulled in as an npm module. So using relative to the file makes this work if you are composing these files. Perhaps a better solution would be to pass in the folder paths as parameters to the construct.

drduhe: Or maybe something like this for ENV detection?
```
export const ROOT_DIR = process.env.APP_ROOT
  ? path.resolve(process.env.APP_ROOT)
  : path.resolve(__dirname);
  ```

lib/user-interface/userInterfaceConstruct.ts
bedanley: Updated to use native function

lib/docs/docConstruct.ts
bedanley: Consolidate the paths to single file.

lib/core/apiBaseConstruct.ts
bedanley: Added

lib/core/apiDeploymentConstruct.ts
bedanley: added

PR #299, opened 2025-04-17
Use ESbuild for deployer lambdas

bedanley:

Migrate ecs models and vector store deployers to be nodejs lambdas using esbuild bundling. ESBuild will pull in all dependencies, which reduces the size of the lambda so that it can be deployed using a native NodeJS runtime, instead of a Docker container. 


drduhe (commented):
LGTM

estohlmann (approved):

PR #303, opened 2025-04-25
Support Image Generation

estohlmann:
*Description of changes:*

Added support for image generation models meeting the following core features:
- Added 3rd supported model type: **IMAGEGEN**
  - This can be added via API and UI
- Customers can view images generated in their convo
- Customers can download and copy each individual image
- Customers can leave a session and come back with their images persisting
- There is a redo button that re-prompts the model to generate the images again with the same settings
- Images can be bulk exported from the session

output


lambda/session/lambda_functions.py
```diff
@@ -87,6 +94,21 @@ def _delete_user_session(session_id: str, user_id: str) -> Dict[str, bool]:
...
+    url: str = s3_client.generate_presigned_url(
+        "get_object",
+        Params={
+            "Bucket": s3_bucket_name,
+            "Key": key,
+            "ResponseContentType": "image/png",
+            "ResponseCacheControl": "no-cache",
+            "ResponseContentDisposition": "inline",
+        },
+        ExpiresIn=3600,
```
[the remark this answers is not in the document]
estohlmann: This is the default for s3, I can remove this I believe and let they system handle that.

lib/user-interface/react/src/components/chatbot/Chat.tsx
```diff
@@ -136,60 +141,119 @@ export default function Chat ({ sessionId }) {
...
+                    const llmClient = createOpenAiClient(chatConfiguration.sessionConfiguration.streaming);
+
+                    // Convert chat history to messages format
+                    let messages = session.history.concat(params.message).map((msg) => ({
+                        role: msg.type === 'human' ? 'user' : msg.type === 'ai' ? 'assistant' : 'system',
+                        content: Array.isArray(msg.content) ? msg.content : selectedModel.modelName.startsWith('sagemaker') ? msg.content :  [{ type: 'text', text: msg.content }]
                     }));
+
+                    const [systemMessage, ...remainingMessages] = messages;
+                    messages = [systemMessage, ...remainingMessages.slice(-(chatConfiguration.sessionConfiguration.chatHistoryBufferSize * 2) - 1)];
```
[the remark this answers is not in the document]
estohlmann: We currently load all messages into context. If we would want to implement pagination, I would suggest doing that as a one off feature but probably a good thing to get in the backlog as a scalability item.

bedanley (dismissed):

lambda/session/lambda_functions.py
```diff
@@ -78,6 +86,8 @@ def _delete_user_session(session_id: str, user_id: str) -> Dict[str, bool]:
     deleted = False
     try:
         table.delete_item(Key={"sessionId": session_id, "userId": user_id})
+        bucket = s3_resource.Bucket(s3_bucket_name)
+        bucket.objects.filter(Prefix=f"images/{session_id}").delete()
```
dustins, line 90: We certainly don't do this everywhere (anywhere?) other than maybe prompt templates to support versioning, but should we consider something like this being configurable and having logical deletes? At some point we'll want auditability most likely and I'm not sure how likely we'll want to delete stuff like this? Probably a bigger discussion as it would affect more than just this.

lambda/session/lambda_functions.py
```diff
@@ -87,26 +97,49 @@ def _delete_user_session(session_id: str, user_id: str) -> Dict[str, bool]:
     return {"deleted": deleted}
 
 
+def _generate_presigned_image_url(key: str) -> str:
+    url: str = s3_client.generate_presigned_url(
+        "get_object",
+        Params={
+            "Bucket": s3_bucket_name,
+            "Key": key,
+            "ResponseContentType": "image/png",
```
dustins, line 106: Is this really always going to be a PNG? Do all models generate PNGs?

lib/user-interface/react/src/components/utils.ts
```diff
@@ -78,3 +78,44 @@ export const getDisplayableMessage = (content: MessageContent, ragCitations?: st
...
+
+        for (let j = 0; j < slice.length; j++) {
+            byteNumbers[j] = slice.charCodeAt(j);
+        }
+
+        const byteArray = new Uint8Array(byteNumbers);
+        byteArrays.push(byteArray);
+    }
+
+    return new Blob(byteArrays, { type: mimeType });
```
dustins, line 120: Why are we doing this in 512 chunks? Since we have the whole base64 string from the start I think this can be simplified...

```typescript
function base64ToBlob(base64: string, mimeType = ''): Blob {
    const byteCharacters = atob(base64);
    const byteNumbers = new Array(byteCharacters.length);

    for (let i = 0; i < byteCharacters.length; i++) {
        byteNumbers[i] = byteCharacters.charCodeAt(i);
    }

    const byteArray = new Uint8Array(byteNumbers);
    return new Blob([byteArray], { type: mimeType });
}
```

lib/user-interface/react/src/components/model-management/create-model/BaseModelConfig.tsx
```diff
@@ -57,20 +59,21 @@ export function BaseModelConfig (props: FormProps<IModelRequest> & BaseModelConf
...
+                        if (fields.modelType === ModelType.embedding || fields.modelType === ModelType.imagegen) {
                             fields['streaming'] = false;
                         }
 
-                        // turn off summarization for embedded models
-                        if (fields.modelType === ModelType.embedding && props.item.features.includes('summarization')) {
-                            fields['features'] = props.item.features.filter((feature) => feature !== 'summarization');
+                        // turn off summarization and image input for embedded and imagegen models
+                        if ((fields.modelType === ModelType.embedding || fields.modelType === ModelType.imagegen)) {
+                            fields['features'] = props.item.features.filter((feature) => feature.name !== 'summarization' && feature.name !== 'imageInput');
```
dustins, line 68: Do we not have an enum of our supported features? Shame on us if not!

lib/user-interface/react/src/components/chatbot/Message.tsx
```diff
@@ -17,47 +17,122 @@
...
+                                    }
+                                }}
+                            />
+                        </Grid>;
                 }
                 return null;
             });
         }
-        return <div>{content}</div>;
+        return message.type === 'human' ?
```
dustins, line 121: Seems like we do a lot of this type of thing in a few places. Maybe we should have a `isHumanMessage` function which would be cleaner, easier to read, and have the magic string 'human' in less places. Speaking of...do we have an enum for the types?

lib/user-interface/react/src/components/chatbot/Message.tsx
```diff
@@ -17,47 +17,122 @@
...
+                                            ...chatConfiguration,
+                                            sessionConfiguration: {
+                                                ...chatConfiguration.sessionConfiguration,
+                                                imageGenerationArgs: {
+                                                    size: metadata?.imageGenerationParams?.size,
+                                                    numberOfImages: metadata?.imageGenerationParams?.n,
+                                                    quality: metadata?.imageGenerationParams?.quality
+                                                }
+                                            }
+                                        });
```
dustins, line 110: We do this 3x places just in this file, and a bunch of other places as well. Is this something we can genericize, maybe using lodash merge or apply?

lambda/session/lambda_functions.py
estohlmann: Yes, models generate PNGs. Ironically I started off as JPEG and it worked for everything other than copy to clipboard. Unfortunately you can only copy pngs which is why it is hardcoded.

lib/user-interface/react/src/components/chatbot/Message.tsx
estohlmann: I updated to use an ENUM. I looked in this file, it is only used twice. I almost think that it makes sense to do these explicit checks vs maintaining 3 (soon to be 5) different booleans. 

- Human
- AI
- System

Coming eventually:
- Tool
- Tool response

lib/user-interface/react/src/components/model-management/create-model/BaseModelConfig.tsx
estohlmann: No we didn't, fixed!

dustins (approved):

PR #305, opened 2025-04-30
Directive prompt templates

estohlmann:
*Description of changes:*

We previously supported Persona Prompt Templates. This PR adds the ability for users to create/share directive based templates and pull them in to your chat sessions.


batzela (approved):
Great work!

lambda/session/lambda_functions.py
```diff
@@ -95,17 +117,23 @@ def list_sessions(event: dict, context: dict) -> List[Dict[str, Any]]:
...
-                "sessionId": session["sessionId"],
-                "firstHumanMessage": next(
-                    (msg["content"] for msg in session.get("history", []) if msg.get("type") == "human"), ""
-                ),
-                "startTime": session.get("startTime", None),
-                "createTime": session.get("createTime", None),
-            }
-        )
+
+    with ThreadPoolExecutor() as executor:
```
dustins, line 121: FWIW I think it is more common to reuse the same `ThreadPoolExecutor`. So having something like this at the top of the file and then not creating a new one every time. Then you wouldn't have to have these nested methods.

```python
executor = ThreadPoolExecutor(max_workers=10)
```

lambda/session/lambda_functions.py
```diff
@@ -142,11 +194,65 @@ def delete_user_sessions(event: dict, context: dict) -> Dict[str, bool]:
...
+            return {"statusCode": 400, "body": json.dumps({"error": f"Invalid JSON: {str(e)}"})}
+
+        if "message" not in body:
+            return {"statusCode": 400, "body": json.dumps({"error": "Missing required fields: messages"})}
+
+        message = body["message"]
+
+        if (
+            message.get("type", None) == "image_url"
+            and message.get("image_url", {}).get("url", None) is not None
```
dustins, line 230: `message.get("image_url", {})` is repeated quite a bit in this function. I think pulling out some of the more common things up after `message = body["message"]` would clean up the code and make this if statement easier to understand.

lib/chat/api/session.ts
```diff
@@ -150,6 +171,19 @@ export class SessionApi extends Construct {
...
+                name: 'attach_image_to_session',
+                resource: 'session',
+                description: 'Attaches image to session',
+                path: 'session/{sessionId}/attachImage',
+                method: 'PUT',
+                environment: {
+                    SESSIONS_TABLE_NAME: sessionTable.tableName,
+                    SESSIONS_BY_USER_ID_INDEX_NAME: byUserIdIndex,
+                    GENERATED_IMAGES_S3_BUCKET_NAME: imagesBucket.bucketName
                 },
```
dustins, line 187: Are these same 3 env variables are repeated for every lambda definition here? Should _somebody_ turn this into a variable that can be reused? 😇

lib/user-interface/react/src/components/model-management/create-model/BaseModelConfig.tsx
```diff
@@ -124,11 +125,11 @@ export function BaseModelConfig (props: FormProps<IModelRequest> & BaseModelConf
                             props.setFields({'streaming': detail.checked})
                         }
                         onBlur={() => props.touchFields(['streaming'])}
-                        disabled={props.item.modelType === ModelType.embedding}
+                        disabled={props.item.modelType === ModelType.embedding || props.item.modelType === ModelType.imagegen}
```
dustins, line 128: I see this a few times and it would be easier to understand and cleaner to pull this out into a reusable variable.

dustins (dismissed):

dustins (approved):

dustins (approved):

PR #338, opened 2025-07-01
MCP Guardrails

estohlmann:
*Description of changes:*
Added the following to support guardrails:
- Added stop button so user can break the AI/Tool Call loop
- Added infinite loop prevention, requiring user feedback after 20 continuous AI/Tool calls
- Defaulting into "Safe Mode" requiring user approval prior to calling tools (overrides coming next PR)

image

image

### DIFF
--- a/lib/user-interface/react/src/components/chatbot/Chat.tsx
+++ b/lib/user-interface/react/src/components/chatbot/Chat.tsx
@@ -69,9 +69,9 @@ import { getButtonItems, useButtonActions } from './config/buttonConfig';
 import { useListMcpServersQuery } from '@/shared/reducers/mcp-server.reducer';
 import { setConfirmationModal } from '@/shared/reducers/modal.reducer';
 import ConfirmationModal from '@/shared/modal/confirmation-modal';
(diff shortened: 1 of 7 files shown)


edparris (approved):

lib/user-interface/react/src/components/chatbot/Chat.tsx
```diff
@@ -298,10 +351,34 @@ export default function Chat ({ sessionId }) {
         }
     }, [session]);
 
+    // Reset tool call counter when session changes
+    useEffect(() => {
+        consecutiveToolCallCount.current = 0;
```
jmharold, line 356: NB: We will want to ensure sessions are terminated when a user switches chats. Since tool calls are being reset on session change, the (hopefully unlikely) scenario exists in which a user has no-approval (YOLO) mode enabled, swaps sessions, gets distracted, and the chatbot makes more tool calls that anticipated.

estohlmann: Good call out, we need to get a story created in the backlog to address session switch and we can make sure this addressed then!

PR #365, opened 2025-08-19
Feature/enhanced model mgmt

jmharold:
*Description of changes:*
- Added a `Model Library` which non-admin users can utilize to view models they have access to
  - Admins have the ability to turn this feature off from the `Configuration` panel.
- Added `Model Description` field to models
- Added ability to view model details post creation. To do this Admins can select their model (LISA or 3rd Party) from the model configuration page -> Actions -> Update and be able to click through all pages and view their models settings / fields
- Added model restriction. Admins can lock models down by groups. Users who are not members of the groups associated with a model will not be able to select the model when starting a chat or view it in the model library
  - If a group is not assigned to a model or all groups are removed from a model it will be defaulted to `(Public)` which means all users will be able to access the model
- Added the ability to update certain fields post creation
  - 3rd Party hosted models can update: `Model Description`, `Model Features`, and `Allowed Groups`
  - LISA Hosted models can update all the above and: `Shared Memory Size` of the Container, `Container Health Check Config`, `Container Environment`, and `Auto Scaling Capacity` (min/max/desired capacity, cooldown, default instance warmup).

  
*Notes:*
- When updating the Container Configuration a container restart will be necessary (causing a temporary outage for that model) and Admins will need to acknowledge a warning before being allowed to submit the update form. If multiple instances are spun up it is possible that ECS will instead perform a rolling deployment and there will not be downtime. Either way the acknowledgement is in place for potential outage / degradation of service 
- When adding toolcalls to a model there may be issue with models that were not originally configured as tool calling models due to missing parameters that litellm is expecting. Additionally, a new chat will need to be started for the model to acknowledge the fact that it now has access to tools.

### DIFF
--- a/lib/user-interface/react/src/components/model-management/create-model/CreateModelModal.tsx
+++ b/lib/user-interface/react/src/components/model-management/create-model/CreateModelModal.tsx
@@ -187,61 +187,75 @@ export function CreateModelModal (props: CreateModelModalProps) : ReactElement {
                 'autoScalingConfig'
             ]);
 
-            const updateRequest: any = {
-                modelId: props.selectedItems[0].modelId,
-                ...(updateFields.streaming !== undefined && { streaming: updateFields.streaming }),
-                ...(updateFields.enabled !== undefined && { enabled: updateFields.enabled }),
-                ...(updateFields.modelType !== undefined && { modelType: updateFields.modelType }),
-                ...(updateFields.modelDescription !== undefined && { modelDescription: updateFields.modelDescription }),
-                ...(updateFields.allowedGroups !== undefined && { allowedGroups: updateFields.allowedGroups }),
-                ...(updateFields.features !== undefined && { features: updateFields.features }),
-                ...(updateFields.containerConfig !== undefined && {
-                    containerConfig: {
-                        ...updateFields.containerConfig,
-                        environment: (() => {
-                            const originalEnv = props.selectedItems[0]?.containerConfig?.environment || {};
-                            const result: any = {};
-                            const currentKeys = new Set<string>();
+            // Build the update request
+            const baseRequest = { modelId: props.selectedItems[0].modelId };
 
-                            // Add/update current variables and track keys
-                            (state.form.containerConfig.environment || []).forEach(({key, value}: any) => {
-                                if (key && key.trim() !== '') {
-                                    result[key] = value;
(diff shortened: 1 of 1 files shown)


lambda/models/handler/get_model_handler.py
```diff
@@ -23,9 +25,37 @@
...
+        model = to_lisa_model(ddb_item)
+
+        # Check if user has access to this model based on groups
+        if not is_admin and user_groups is not None:
+            if not self._user_has_group_access(user_groups, model.allowedGroups or []):
+                raise ModelNotFoundError(f"Model '{model_id}' was not found.")
+
+        return GetModelResponse(model=model)
+
+    def _user_has_group_access(self, user_groups: List[str], allowed_groups: List[str]) -> bool:
```
estohlmann, line 45: Do we not have a method for this somewhere in utils? If we don't we should add this because this is going to be a thing for MCP. We would currently do this for:
- Models
- Prompt Templates
- Vector Stores

lib/user-interface/react/src/components/model-management/create-model/CreateModelModal.tsx
```diff
@@ -110,26 +146,103 @@ export function CreateModelModal (props: CreateModelModalProps) : ReactElement {
...
+                'modelDescription',
+                'allowedGroups',
+                'features',
+                'containerConfig',
+                'autoScalingConfig'
+            ]);
+
+            const updateRequest: any = {
+                modelId: props.selectedItems[0].modelId,
+                ...(updateFields.streaming !== undefined && { streaming: updateFields.streaming }),
```
estohlmann, line 192: All of these spreads are kind of gross, I think lodash has a merge operator that we can use. Might be worth exploring. I am not sure if it has the "smarts" to ignore undefined but I have to think that is possible

estohlmann: I would research it for 5 minutes, if you can't find a solution we can ship it

jmharold: Addressed in most recent commit. Spoke with bedanley about how I can break this up further in a future commit to better align the model updating workflow with SOLID principals / Tiered architecture

bedanley (approved):

PR #383, opened 2025-09-04
UI Enhancements

jmharold:
*Description of changes:*
- Removed a tailwind import that was resetting browser formatting. Markdown will now render properly in browsers.
- Added formatting for inline-code and blocks of text (code blocks with no language specifier)
- LISA can now render Mathematic equations and symbols using LaTex syntax
- LISA can now render Mermaid digrams.
  - The code for these diagrams can be copied
  - These diagrams can also be downloaded as pngs
- Added ability to copy user message
- Changed session redirect behavior when deleting a session
  - When deleting the session you are in, you will create a new session using the old session ID instead of being directed to the home page
  - When deleting a session other than the one your are in, you will be not be returned to the home page and your screen will not force a reload.
- Screen now follows the stream as response is being generated
- Example stop words have been removed (variations of "User")
- Behavior of Adding / Removing stop words has been changed
  - Previous behavior was that you needed to send a message to have changes to stop words persist
  - Current behavior is that we will write changes immediately after they are made in the UI
- Added stacked notifications
- Updated default System Prompt (persona) with more robust specific instructions / persona.
  - **NOTE:** There are two pieces in here important for anyone wanting to overwrite with their own prompt. If you are using mermaid or the math rendering, you should copy the paragraphs from the default system prompt telling the model how to use the tools to your own system prompt.
- Stacked Session History

Screenshot 2025-09-04 at 2 23 34 PM
Screenshot 2025-09-04 at 3 32 13 PM
Screenshot 2025-09-04 at 3 33 56 PM
Screenshot 2025-09-04 at 3 34 51 PM


lib/user-interface/react/src/components/chatbot/components/Sessions.tsx
```diff
@@ -17,6 +17,7 @@
 import SpaceBetween from '@cloudscape-design/components/space-between';
 import Link from '@cloudscape-design/components/link';
 import Header from '@cloudscape-design/components/header';
+import ExpandableSection from '@cloudscape-design/components/expandable-section';
 import { ButtonDropdown, Input, Popover, Modal, FormField, Grid } from '@cloudscape-design/components';
```
bedanley, line 21: Can be imported in the next line

bedanley (approved):

PR #404, opened 2025-09-30
Fix/Workshop Bugs

estohlmann:
*Description of changes:*
- We currently can't deploy two LISA's in the same account due to metrics stack naming
- Bugs around preferred UserName and MCP connections
- MCP workbench is created in inactive state


lambda/mcp_server/lambda_functions.py
```diff
@@ -178,19 +178,38 @@ def _is_member(user_groups: List[str], prompt_groups: List[str]) -> bool:
     return bool(set(user_groups) & set(prompt_groups))
 
 
+def _set_can_use(
+    connections: Dict[str, Any], user_id: Optional[str] = None, groups: Optional[List[str]] = None
+) -> Dict[str, Any]:
+    items = connections.get("Items", [])
+    formatted_groups = [f"group:{group}" for group in groups]
```
dustins, line 185: Won't this fail is groups is `None`? Maybe groups should default to an empty list instead of None, then you don't need to do that check a few lines down see if `groups` is truth

lambda/mcp_server/lambda_functions.py
```diff
@@ -178,19 +178,40 @@ def _is_member(user_groups: List[str], prompt_groups: List[str]) -> bool:
     return bool(set(user_groups) & set(prompt_groups))
 
 
+def _set_can_use(
+    connections: Dict[str, Any], user_id: Optional[str] = None, groups: Optional[List[str]] = None
+) -> Dict[str, Any]:
+    if groups is None:
+        groups = []
```
dustins, line 185: Can't we just do `Optional[List[str]] = []` and drop that if statement?

PR #522, opened 2025-10-21
Feature/bedrock guardrails state machines

jmharold:
*Description of changes:*
- Update `create`, `update`, and `delete` state machine to support guardrails.
- Fix for `litellm_client` in which we were expecting a body for HTTP Requests which could be body-less.

### DIFF
--- a/lib/models/state-machine/delete-model.ts
+++ b/lib/models/state-machine/delete-model.ts
@@ -183,7 +183,6 @@ export class DeleteModelStateMachine extends Construct {
 
         // State Machine definition
         setModelToDeleting.next(deleteFromLitellm);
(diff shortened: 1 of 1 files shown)


lib/models/state-machine/delete-model.ts
```diff
@@ -164,7 +183,9 @@ export class DeleteModelStateMachine extends Construct {
 
         // State Machine definition
         setModelToDeleting.next(deleteFromLitellm);
-        deleteFromLitellm.next(deleteStackChoice);
+        // deleteFromLitellm.next(deleteStackChoice);
```
estohlmann: remove

estohlmann (approved):

PR #627, opened 2025-12-17
Bug/apply utc time

bedanley:
Apply consistant dates for object creation and UI rendering. Dates were not created consistently in python, which was not storing the timezone with iso timestrings. Adding a default UTC timezone, and updating factory methods to create timestamps using UTC timezone. Where possible, the values have been updated to use epoch mills, instead of timestrings. 


nishi7409 (changes requested):
Looks good, could you please rerun precommit & testing commands. Can merge after

PR #637, opened 2025-12-22
Bug/session history

bedanley:
Addressing session history not loading when selected. Also, cleaning up react form warnings. 

### DIFF
--- a/lib/user-interface/react/src/components/chatbot/components/Message.tsx
+++ b/lib/user-interface/react/src/components/chatbot/components/Message.tsx
@@ -216,12 +216,12 @@ export const Message = React.memo(({ message, isRunning, showMetadata, isStreami
         },
     }), [isStreaming, onMermaidRenderComplete]); // Include isStreaming and onMermaidRenderComplete so the component can access them
 
-    const renderContent = (messageType: string, content: MessageContent, metadata?: LisaChatMessageMetadata) => {
+    const renderContent = (content: MessageContent, metadata?: LisaChatMessageMetadata) => {
         if (Array.isArray(content)) {
-            return content.map((item, index) => {
-                if (item.type === 'text') {
-                    return item.text.startsWith('File context:') ? <div key={index}></div> : <div key={index}>{getDisplayableMessage(item.text, message.type === MessageTypes.AI ? ragCitations : undefined)}</div>;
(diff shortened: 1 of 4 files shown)


lib/user-interface/react/src/components/chatbot/components/Message.tsx
```diff
@@ -220,7 +220,7 @@ export const Message = React.memo(({ message, isRunning, showMetadata, isStreami
         if (Array.isArray(content)) {
             return content.map((item, index) => {
                 if (item.type === 'text') {
-                    return item.text.startsWith('File context:') ? <></> : <div key={index}>{getDisplayableMessage(item.text, message.type === MessageTypes.AI ? ragCitations : undefined)}</div>;
+                    return item.text.startsWith('File context:') ? <div key={index}></div> : <div key={index}>{getDisplayableMessage(item.text, message.type === MessageTypes.AI ? ragCitations : undefined)}</div>;
```
dustins: Looks like we're repeating ourselves, so maybe just something like below?

```suggestion
                    return <div key={index}>{ item.text.startsWith('File context:') && getDisplayableMessage(item.text, message.type === MessageTypes.AI ? ragCitations : undefined)}</div>;
```

bedanley: Fixed to not return empty element

lib/user-interface/react/src/components/chatbot/components/Sessions.tsx
```diff
@@ -155,7 +154,10 @@ export function Sessions ({ newSession }) {
             // Reset the tracking state
             setSessionBeingDeleted(null);
         } else if (!isDeleteByIdLoading && isDeleteByIdError) {
-            notificationService.generateNotification(`Error deleting session: ${deleteByIdError.data?.message ?? deleteByIdError.data}`, 'error');
+            const errorMessage = 'data' in deleteByIdError && deleteByIdError.data
+                ? (typeof deleteByIdError.data === 'string' ? deleteByIdError.data : deleteByIdError.data.message || 'Unknown error')
+                : 'Unknown error';
+            notificationService.generateNotification(`Error deleting session: ${errorMessage}`, 'error');
```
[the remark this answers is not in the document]
bedanley: Moved to reducer

PR #647, opened 2026-01-04
Feature/cypress tests

bedanley:
Add additional Cypress smoke tests:
- Ensure all admin pages load with data
- Ensure chat prompts render responses
- Ensure chat sessions are selectable and load properly
- Ensure non-admins can't navigate to admin pages
- Refactored tests to be sharable with E2E suite. 

Also address a small bug fix to make mcp-workbench redirect like other admin pages. 

### DIFF
--- a/lib/docs/.vitepress/config.mts
+++ b/lib/docs/.vitepress/config.mts
@@ -56,7 +56,7 @@ const navLinks = [
       { text: 'Model Compatibility', link: '/config/model-compatibility' },
       { text: 'Model Management API', link: '/config/model-management-api' },
       { text: 'Model Management UI', link: '/config/model-management-ui' },
-      { text: 'Guardrails', link: '/config/guardrails' },
+      { text: 'Bedrock Guardrails', link: '/config/guardrails' },
       { text: 'Usage & Features', link: '/config/usage' },
       { text: 'RAG Repository', link: '/config/repositories' },
       { text: 'Langfuse Tracing', link: '/config/langfuse-tracing'},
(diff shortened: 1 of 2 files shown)


cypress/cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "Using fixtures to represent data",
```
estohlmann: do we want this?

cypress/cypress/support/commands.ts
```diff
@@ -0,0 +1,53 @@
...
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+
+/// <reference types="cypress" />
+// ***********************************************
+// This example commands.ts shows you how to
```
estohlmann: seems like we can remove these example files that we aren't using

cypress/src/smoke/fixtures/session-detail.json
```diff
@@ -0,0 +1,100 @@
+{
+  "sessionId": "f56fc284-629c-4ba7-ab3d-56f4a21c13ee",
```
estohlmann: What's the difference between this fixture and chat-session.json?

cypress/src/support/adminHelpers.ts
```diff
@@ -56,3 +66,136 @@ export function checkNoAdminButton () {
...
+            cy.get('tbody tr, [role="row"]:not([role="columnheader"])')
+                .should('have.length.at.least', minRows);
+        });
+}
+
+/**
+ * Verify that a Cloudscape table component has rendered with data
+ * Uses Cloudscape-specific selectors for better reliability
+ */
+export function verifyCloudscapeTableHasData (minRows: number = 1) {
```
estohlmann: Future TODO: We should probs move this to more of a utility package vs in the adminHelpers since this isn't specific to admin functionality and generic app functionality handler logic

cypress/cypress/fixtures/example.json
bedanley: Removed folder

PR #706, opened 2026-01-29
Feature/abstract auth

bedanley:
This is a mixed bag of features and bugs found during deployment in isolated regions.

Features:
- Add auth abstraction - allows user to override authentication mechanisms with minimal changes
- Update asset builder to use containers - now python requirements are correctly applied to the lambda runtime instead of using local os / python version binaries
- Explicitly set the network mode of ECS containers - VPC for rest-api/mcp workbench, and bridge for daemon models
- Add additional JWT supported algos
- Add default logging buckets for lambdas with retention policy
- Simplified cdk layer requirements to only include needed deps for lambdas

Bugs:
- properly set admin endpoints for configuration
- Set default UI configuration based on deployed components
- Only render RAG in UI if RAG is deployed
- Set protection policy to match deletion policy 
- Add missing cdk layer package - constructs
- Correct stack dependency order
- Block public s3 buckets by default
- Revert litellm table name change causing recreation
- Properly set the platform with image creation to allow deployments from not AL2 environments

### DIFF
--- a/lib/serve/rest-api/src/api/endpoints/v2/litellm_passthrough.py
+++ b/lib/serve/rest-api/src/api/endpoints/v2/litellm_passthrough.py
@@ -63,6 +63,8 @@ OPENAI_ROUTES = (
     # Create images
     "images/generations",
     "v1/images/generations",
+    "images/edits",
+    "v1/images/edits",
     # Audio routes
     "audio/speech",
     "v1/audio/speech",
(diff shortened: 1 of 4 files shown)


lib/docs/config/model-management-api.md
```diff
@@ -32,7 +32,7 @@ curl -s -H "Authorization: Bearer <admin_token>" -X GET https://<apigw_endpoint>
       },
       "containerConfig": {
         "image": {
-          "baseImage": "vllm/vllm-openai:v0.5.0",
+          "baseImage": "vllm/vllm-openai:latest",
```
batzela: Consider pinning this to a version hash.

lib/docs/config/model-management-api.md
```diff
@@ -118,7 +118,7 @@ POST https://<apigw_endpoint>/models
   "streaming": true,
   "containerConfig": {
     "image": {
-      "baseImage": "vllm/vllm-openai:v0.5.0",
+      "baseImage": "vllm/vllm-openai:latest",
```
batzela: Here too.

estohlmann (approved):

mcp_server_deployer/src/lib/scripts/astral-install.sh
```diff
@@ -0,0 +1,2080 @@
+#!/bin/sh
```
Ernest-Gray: This is a very large script.  I suggest splitting it into seperate scripts if logical and valid then having docstrings/a README to go along with the files.

PR #720, opened 2026-02-04
Feature/typesafe sessions

bedanley:
Add typesafe objects to session lambdas. Malformed requests should now be caught by pytdantic and returned to the user as a 400 error. 

We are also adding additional validation to rest-api with null byte protection and header sanitation. 


batzela (dismissed):
Looks good to me.

PR #762, opened 2026-02-17
Fix/stale redux cache

gingerknight:
### Problem
Users on long-lived or resumed sessions experienced failures connecting to models and loading sessions. Investigation confirmed that redux-persist was writing all RTK Query caches to localStorage and rehydrating them on every page load, causing the app to serve stale data from previous sessions instead of fetching fresh state from the API.

### Solution
redux-persist was removed entirely. The RTK Query slices already have refetchOnFocus and refetchOnMountOrArgChange configured and handle their own cache freshness correctly without persistence. User identity state is now populated directly from the live OIDC session via a new UserStateSync component inside AuthProvider, which fires on both fresh login and page refresh. The login modal was also updated to respect auth.isLoading to prevent a brief flash during OIDC session resolution.

Verified manually: hard refresh retains access, sign out/in flow works, admin routes gated correctly, models and sessions load fresh on every page load.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -52,7 +52,7 @@ repos:
   hooks:
     - id: codespell
       entry: codespell
-      args: ['--skip=*.git*,*cdk.out*,*venv*,*mypy_cache*,*package-lock*,*node_modules*,*dist/*,*poetry.lock*,*coverage*,*models/*,*htmlcov*,*TIKTOKEN_CACHE/*,*test/cdk/stacks/__baselines__/*', "-L=xdescribe,assertIn,afterAll"]
+      args: ['--skip=*.git*,*cdk.out*,*venv*,*mypy_cache*,*package-lock*,*node_modules*,*dist/*,*/public/*,*poetry.lock*,*coverage*,*models/*,*htmlcov*,*TIKTOKEN_CACHE/*,*test/cdk/stacks/__baselines__/*', "-L=xdescribe,assertIn,afterAll"]
       pass_filenames: false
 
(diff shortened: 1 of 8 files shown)


gingerknight:
Did some console testing to validate the cache values are cleared and the user field are preserved

```js
// Corrupt the models cache with a fake stale model
const cache = JSON.parse(localStorage.getItem('persist:lisa'));
const models = JSON.parse(cache.models);
models.queries['getAllModels(undefined)'] = {
  status: 'fulfilled',
  endpointName: 'getAllModels',
  data: [{ modelId: 'STALE_FAKE_MODEL', modelName: 'Stale Model From Old Session' }],
  fulfilledTimeStamp: 0
};
cache.models = JSON.stringify(models);
localStorage.setItem('persist:lisa', JSON.stringify(cache));
console.log('Injected stale model:', models.queries['getAllModels(undefined)'].data);

```

```js
[
  {
    "modelId": "STALE_FAKE_MODEL",
    "modelName": "Stale Model From Old Session"
  }
]
```

```js
location.reload()

const cache = JSON.parse(localStorage.getItem('persist:lisa'));
const models = JSON.parse(cache.models);

models.queries['getAllModels(undefined)'].data
//->undefined
```

dustins:
Do you really need to use `redux-persist`? What data is saved from being loaded again that we don't want refreshed after a user comes back? This seems like a reasonable fix for the problem but is the thing causing the problem really buying anything you can't get with configuring RTK data caching lifetimes? MLSpace has a similar unnecessary redux-persist usage too.

PR #935, opened 2026-04-08
Update production settings in LiteLLM

bedanley:
Update configuration and deployment of LiteLLM to adhere to best practices. Changes include:
- Enable "production" mode
- Enable DB migration using Prisma
- Remove verbose logs by default
- Generate prisma client before threading workers
- Update Threads / workers to match vCPU instance type - m5.xl = 4vCPU = 4 Total threads - 2xLisa http/auth 2xLitellm Workers
- Enable retrys with db connections
- Set max connections 
- Clean up dead model config generation


lib/docs/package.json
```diff
@@ -16,7 +16,7 @@
   "author": "",
   "license": "Apache-2.0",
   "devDependencies": {
-    "vitepress": "^1.6.4",
+    "vitepress": "^2.0.0-alpha.17",
```
estohlmann, line 19: not sure we really should be using an alpha release

bedanley: This was only for doc generation. Less of an issue. The latest version is needed to support newer versions of vite and esbuild.

PR #945, opened 2026-04-10
fix: updating mcp tool registration logic

drduhe:
*Issue #, if available:*

## Summary

Improves **MCP Workbench tool lifecycle** (S3 → in-VPC HTTP rescan, FastMCP `remove_tool` / registry updates, shared server id) and keeps the **chat UI** in sync with workbench tool files using stock `use-mcp`. **Amazon Bedrock Agents** stay available in chat alongside MCP: merged OpenAI tool definitions, `invokeBedrockAgent` routing, approval flow, and footer banner counts.

Functionally, this makes tools register and de-register from the MCP Workbench near instantaneously rather than taking 5-7 minutes per interaction. This is due to not having to spin up new ECS tasks every time a tool is added/deleted/updated via the workbench. 

## Backend / infrastructure

- **`lambda/mcp_workbench/lambda_functions.py`** — Loads `mcpWorkbenchServerId` from `mcp_workbench_server_id.json` (single source of truth with the UI).
- **`lambda/mcp_workbench/s3_event_handler.py`** — Replaces ECS forced redeploy with **`GET`** to the workbench rescan endpoint (default path `v2/mcp/rescan`), optional **management bearer** from Secrets Manager, configurable **`WORKBENCH_RESCAN_DELAY_SECONDS`**, and avoids importing the Common-layer-only `utilities.common_functions` chain from the bare Lambda asset.
- **`lib/serve/mcpWorkbenchConstruct.ts`** — S3 event Lambda: VPC, lambda SG, common layer, endpoint/rescan env vars, Secrets Manager IAM; drops ECS/SSM permissions used for forced redeploy.
- **`lib/serve/mcp-workbench/s6-overlay/.../s3mount/run`** — Configurable **`RCLONE_POLL_INTERVAL`** (default `30s`) so S3 listing changes surface in the mount before rescan.

## MCP Workbench service (Python)

- **`mcp_server.py`** — On rescan: `remove_tool` for removed/updated tools when supported; `update_registry`; timezone-aware UTC timestamps.
- **`auth.py`** — Expired OIDC ID tokens logged at debug; other JWT issues as warning (less noisy than full exceptions).
- **`README.md`** — Documents S3 → Lambda → rescan and rclone polling; minor doc formatting.

## UI (React)

- **`mcpWorkbenchServerId.ts`** + **`mcp_workbench_server_id.json`** — Same UUID for the auto-registered workbench server in Lambda and UI.
- **`utils.ts`** — `isWorkbenchMcpServer()` (id match, then origin vs `MCP_WORKBENCH_URI`).
- **`mcp.hooks.tsx`** — Stable `useMcp` options (header / `clientConfig` fingerprints), **`autoRetry: false`**, aggregates tools when `ready` and clears when not, prunes state when servers are disabled, **`readyMcpServerCount`**. For **workbench** servers, the connection **`key`** includes a **`workbenchToolListFingerprint`** so when the tool-file list from the API changes, the MCP client **remounts** and **`use-mcp` runs `tools/list` again** (upstream `use-mcp` does not expose a refresh API).
- **`Chat.tsx`** — `useListMcpToolsQuery` drives that fingerprint (file ids + `updated_at` / `size` so edits matter). **`useMultipleMcp`** receives `{ workbenchToolListFingerprint }` only. **Bedrock:** `useListBedrockAgentsQuery` / `useInvokeBedrockAgentMutation`, merged **`openAiTools`** when MCP and/or Bedrock components are enabled, `callToolWithBedrock`, `toolToServerMapMerged`, Bedrock-aware **`useToolChain`** props, confirmation modal, and footer banner.
- **`McpWorkbenchManagementComponent.tsx`** — Default editor sample as a template literal (readability).
- **`McpServerDetails.tsx`** — Optional chaining on **`disabledTools`**.

**Dependency note:** There is **no** `patches/use-mcp+…` (or other) **`patch-package`** patch for MCP in this PR; behavior relies on **remount + fingerprint** instead of forking `node_modules`.

## Developer experience: `scripts/run-pytest.sh`
Adds a small **pytest launcher** so `npm run test:lambda`, `test:mcp-workbench`, `test:python:coverage`, and related scripts use a **predictable interpreter**:
- If **`PYTHON`** is set, runs `"$PYTHON" -m pytest …`.
- Else if **`./.venv/bin/python`** exists and is executable, uses that (matches “activated venv” behavior without requiring manual `source .venv/bin/activate`).
- Otherwise falls back to **`python3 -m pytest`**.
This keeps local runs aligned with developers who use a repo virtualenv while still working in CI where Python is provided by the image / setup-python. Root **`package.json`** wires the Python test npm scripts through **`bash scripts/run-pytest.sh`** so all of them share the same resolution order.
### `scripts/install-python.sh` and `npm run install:python`
**`install:python`** now runs **`bash scripts/install-python.sh`**, which uses the **same precedence** as `run-pytest.sh` (`PYTHON` → `.venv/bin/python` → `python3`) and installs everything with **`"$PY" -m pip`**, so installs go into the venv when it exists and you avoid mixing system `pip` with a different interpreter.
### `npm run deploy` (`scripts/deploy.mjs`)
Deploy passes a **`repoPythonEnv()`**-derived environment to its child processes: when **`.venv/bin/python`** is present and executable, **`PATH`** is prefixed with **`.venv/bin`** and **`VIRTUAL_ENV`** is set (and path-like **`PYTHON`** still prepends that binary’s directory to **`PATH`**). That way **`npm run build`** steps that shell out to **`python3`**—for example **`bin/package-lambda`** during asset builds—resolve to the repo venv **without** requiring a manual activate, while CI and non-venv setups keep using the default **`python3`** on **`PATH`**.

## Testing / verification

- Workbench: S3 object create/delete → Lambda rescan → tools update without ECS task recycle; with auth, management secret allows authorized rescan.
- UI: Workbench enabled → after tool-file metadata changes (and fingerprint update), MCP connection remounts and the chat tool list matches the server.
- Bedrock: With **Bedrock agents** enabled in config and user/stack preferences, chat still receives Bedrock tools and **`invokeBedrockAgent`** / action-group flows work with the existing approval behavior.


lib/user-interface/react/src/components/mcp-workbench/McpWorkbenchManagementComponent.tsx
```diff
@@ -42,7 +42,150 @@ import ColorSchemeContext from '@/shared/color-scheme.provider';
...
+// Set up a sample tool demonstration that builds the same MCP tool using both
+// the function-based and class-based approaches.
+const DEFAULT_CONTENT = `from mcpworkbench.core.annotations import mcp_tool
```
bedanley, line 47: This is nicer, but it might be work adding it to a dedicated python file that is pulled into the ts code.

bedanley (dismissed):

lib/user-interface/react/src/components/mcp-workbench/McpWorkbenchManagementComponent.tsx
drduhe: Yeah in theory we could cross import it from the MCP workbench module - but that would create a dependency overlap - I will follow up with some further improvements and address this bit down the line - good call out though.

gingerknight (approved):

PR #960, opened 2026-04-15
Bug/update use effect config

bedanley:
Fixed infinite render loops in the React UI by memoizing the authentication user object and refining useEffect dependencies across multiple components. Previously, the auth user object was recreated on every render, causing components that depend on it to continuously re-render and dispatch Redux actions, leading to performance issues and repeated API calls.

### DIFF
--- a/lib/user-interface/react/src/shared/hooks/useAnnouncementNotifier.ts
+++ b/lib/user-interface/react/src/shared/hooks/useAnnouncementNotifier.ts
@@ -14,7 +14,7 @@
   limitations under the License.
 */
 
-import { useEffect, useCallback } from 'react';
+import { useEffect, useCallback, useRef } from 'react';
 import { useAppDispatch } from '@/config/store';
 import { useNotificationService } from '@/shared/util/hooks';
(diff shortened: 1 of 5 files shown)


lib/user-interface/react/src/shared/hooks/useAnnouncementNotifier.ts
```diff
@@ -67,5 +78,5 @@ export function useAnnouncementNotifier (config: IConfiguration | undefined): vo
             true,
             onDismiss,
         );
-    }, [config, clearAnnouncement, notificationService]);
+    }, [config, isEnabled, message, createdAt, clearAnnouncement, notificationService]);
```
gingerknight, line 81: Minor nit: config is still in the dependency array alongside the new const. 
With the guards in place, I _don't think_ this is needed here anymore. We could drop the (!config) early exit since it is already covered by the defaults:
```
isEnabled -> false, 
message -> ''
```

gingerknight (dismissed):
Approved! Good find!

PR #974, opened 2026-04-20
Add Throttling to LISA Serve

estohlmann:
Adding optional throttling to LISA Serve


lib/serve/rest-api/src/main.py
```diff
@@ -82,12 +83,23 @@ async def lifespan(app: FastAPI):  # type: ignore
...
 
 
+@app.middleware("http")
+async def rate_limit(request, call_next):  # type: ignore
+    """Per-user rate limiting middleware.
+
+    Runs after authentication (user identity is available) to enforce
+    per-API-key / per-user request rate limits.
+    """
+    return await rate_limit_middleware(request, call_next)
```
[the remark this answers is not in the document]
estohlmann: We need them to be authed so we can do user specific limit overrides and ensuring that stability is good.

lib/serve/rest-api/src/middleware/rate_limit_middleware.py
```diff
@@ -0,0 +1,280 @@
...
+        if refill_rate <= 0:
+            return False, 60.0
+
+        # How long until one token is available
+        wait = (1.0 - self.tokens) / refill_rate
+        return False, wait
+
+
+# Global bucket store — keyed by user identity string
+_buckets: dict[str, _TokenBucket] = {}
```
bedanley, line 133: Do we have sticky sessions for a user? Otherwise this will need to be synced for each rest-api instance.

bedanley (dismissed):

lib/serve/rest-api/src/middleware/rate_limit_middleware.py
estohlmann: Good call out. Added them to the ALB is rateLimitEnabled is true

bedanley (approved):

PR #993, opened 2026-04-23
refactor: cleanup lambda implementation

drduhe:
## Summary
This PR restructures how Python Lambdas are packaged and how shared code is imported, then threads that model through CDK. It also tightens session list performance, updates the LiteLLM proxy image and request handling, and adds a post-build browser smoke test for the React app.
## Python Lambda layout: `handlers` + `lisa` shared layer
- **Handler code** now lives under `lambda/handlers/<name>/` (per-function zip root). CDK zips each handler directory only (no more monolithic `lambda/` asset for most functions).
- **Shared first-party code** is consolidated under `lambda/shared/python/lisa/` and deployed as a new **`LisaSharedLayer`** (Core stack). At runtime it is available as `import lisa.*` (e.g. `lisa.utilities.*`, `lisa.domain.*`, `lisa.rag.*`, `lisa.mcp.*`, `lisa.session.*`).
- **Imports** across handlers and tests are updated from flat `utilities` / `models` / `repository` / `mcp_server` modules to the `lisa` package namespace.
- **MCP models** move to `lisa.mcp.models`; **MCP workbench server id** is loaded from shared JSON via new `lisa.mcp.workbench` (single source for Lambdas, config, UI).
- **Handler-only modules** are split or renamed for clarity, e.g.:
  - `db_setup_iam_auth` as its own handler package
  - `management_key` as `handler.py` under `handlers/management_key/`
  - `dockerimagebuilder` package exports `handler` for CDK
- **Deployment zip layout:** handlers use `lambda_functions.<fn>` (not `package.subpkg.lambda_functions`) in `registerAPIEndpoint`, matching files at the zip root.
- **Package import fallbacks** where needed (e.g. chat assistant stacks) for “tests as package” vs “Lambda flat zip” import paths.
## CDK: `definePythonLambda`, `lambdaCodeAsset`, `getPythonLambdaLayers`
- New helpers in `lib/util/`:
  - **`definePythonLambda`**: standard Python Lambda defaults (runtime, 180s / 512 MiB, VPC wiring) and auto-attachment of **LisaSharedLayer** (deduped per scope).
  - **`lambdaCodeAsset` / `LAMBDA_HANDLERS`**: maps each API to a single handler directory; supports **`config.lambdaPath` → `.zip`** for air-gapped / prebuilt bundles.
  - **`getPythonLambdaLayers`**: DRYs SSM lookups for third-party layers (common, fastapi, authorizer) with stable construct ids.
- **`registerAPIEndpoint`** now takes **`config`** instead of a string path, uses **`lambdaCodeAsset`**, and **prepends `getLisaSharedLayer`** to the layer list.
- **Authorizer, management key rotation, IAM auth setup, metrics, model/MCP/RAG state machines, MCP workbench, docker image builder**, etc. are migrated from hand-rolled `Function` + `Code.fromAsset(LAMBDA_PATH)` to the new pattern; handler strings are relative to the handler asset (e.g. `state_machine.create_mcp_server.…`).
## Core: `LisaSharedLayer` + SSM
- `CoreConstruct` builds **`lambda/shared/python/`** as a LayerVersion and publishes  
  `${deploymentPrefix}/layerVersion/lisa-shared` in SSM for cross-stack use.
## Session API: performance and IAM
- **`list_sessions`** uses a **DynamoDB projection** that omits large fields (`configuration` / `encrypted_configuration`) to shrink payloads for the session list.
- **Project validation** for `projectId` switches from **BatchGetItem** to **per-id `get_item`**, with **str() normalization** so string and numeric ids match the allowed set; IAM for the session Lambda is updated from **`dynamodb:BatchGetItem`** to **`dynamodb:GetItem`** on the projects table.
- **Configuration Lambda:** when toggling MCP workbench visibility, loads the MCP servers table via **env** instead of importing `mcp_server.lambda_functions` (avoids cross-handler coupling).
## REST API (LiteLLM) container and proxy
- **Dockerfile:** runs **`prisma generate`** against LiteLLM’s bundled schema so the Python Prisma client exists at startup (LiteLLM ≥ 1.83.x may import it lazily; `db push` alone is insufficient).
- **`requirements.txt`:** **LiteLLM `[proxy]` pinned to 1.83.10** (Anthropic Opus 4.7 / adaptive thinking on Bedrock); **boto3** upper bound widened to `<2.0.0` and **click** floor adjusted for resolver compatibility with LiteLLM’s pins.
- **`main.py`:** **CORS middleware re-registered last** so it is **outermost** and adds CORS headers to **all** responses (including auth failures and 5xx), with an explanatory comment on Starlette order.
- **`litellm_passthrough` + `request_utils`:** **`strip_unsupported_model_params`** removes parameters known to be rejected by specific models (e.g. **Opus 4.7 / `top_p`**) using regex on the canonical model path, with logging when stripping occurs.
## React UI
- **`vite.config.ts`:** more aggressive chunking (implied by smoke-test rationale); worth validating no regressions.
- **`SyntaxHighlightedCode`** and updates to **`markdownRenderer`**, **`Chat`**, **`FileUploadModals`**, **`Sessions`**, small **system-banner** tweak.
- **`App.tsx`:** larger routing / bootstrap changes (review for auth, base path, and MCP workbench).
- **Reducers:** small consistent updates (likely loading/error or API path alignment).
- **Removed `vitepress`** from app dependencies; **added Playwright** for smoke tests.
- **New `scripts/smoke-test.mjs` + npm scripts** `test:smoke` / `test:smoke:build`:** headless load of `vite preview` to catch **runtime chunk initialization** errors after build.

### Chat Session + MCP Tools Update
- Reopening a saved chat could **re-run MCP** because the UI treated “last message is assistant + `toolCalls`” as always needing execution, including right after **hydrating** history from the API (`lastProcessedMessageIndex` started at `-1`).
- Sessions could be **saved in a half-finished state** (assistant had requested tools but MCP hadn’t run yet) because the dirty-session effect called **`updateSession` before** the tool chain ran, right after `generateResponse` marked the session dirty.
### Changes
- **Load / replay:** When `loadingSession` goes from `true` → `false`, set `lastProcessedMessageIndex` to the last loaded index so hydrated history is not treated as a new pending tool turn.
- **Persist timing:** Only run the dirty-session `updateSession` when it’s safe: last turn is assistant, not streaming (`!isStreaming`), and **`sessionHistoryHasPendingAssistantToolCalls`** is false (new helper in `sessionPersist.utils.ts`). If we skip because tools are pending, **leave `dirtySession` true** so a later save still runs after tools + follow-up reply.
- **Session settings:** `SessionConfiguration` skips the configuration PUT under the same “pending assistant tool calls” rule so changing options mid–tool-call doesn’t write a bad tail; local config still updates and will be included on the next normal save.
- **Tests:** `sessionPersist.utils.test.ts` covers the helper.

## Tests and tooling
- **`test/lambda`**: mass update of imports and `patch()` targets to `lisa.*` paths; **`conftest` / `pytest.ini` / `pyproject.toml`** adjusted for the new package layout and paths.
## Ops / review notes
- **New handler package:** any future Python API Lambda must be added to **`LAMBDA_HANDLERS`** in `lib/util/lambdaCode.ts` and have a directory under `lambda/handlers/`.
- **CI:** ensure Playwright browsers are installed where `test:smoke` runs, or document skipping.


lib/serve/rest-api/Dockerfile
```diff
@@ -62,6 +62,20 @@ RUN mkdir -p /root/.cache && \
...
     fi
 
+# Generate the Prisma Python client against LiteLLM's bundled schema.
+# LiteLLM's runtime `--use_prisma_db_push` flag only runs `prisma db push`
+# (schema sync); it does NOT run `prisma generate` (Python client codegen).
+# Starting with litellm[proxy] >= 1.83.x, the proxy imports `from prisma import
+# Prisma` lazily at startup and raises "The Client hasn't been generated yet"
+# if the Python client was never generated. The schema lives inside the
+# litellm_proxy_extras package (e.g. .../site-packages/litellm_proxy_extras/schema.prisma).
+RUN python -c "import importlib.util, pathlib; spec = importlib.util.find_spec('litellm_proxy_extras'); print(pathlib.Path(spec.origin).parent / 'schema.prisma')" > /tmp/litellm_schema_path \
```
bedanley, line 72: Adding a note that we want to double check this with some different builds in VPC without internet connection.

lib/serve/rest-api/src/main.py
```diff
@@ -130,6 +139,18 @@ async def security_check(request, call_next):  # type: ignore
     return await security_middleware(request, call_next)
 
 
+# Enable CORS. This MUST be the last middleware registered so it becomes the
+# outermost wrapper and can attach CORS headers to every response, including
+# error responses produced (or short-circuited) by the middleware below.
+app.add_middleware(
```
bedanley, line 147: We will have a conflict with the security PR adding CORS settings.

lib/user-interface/react/src/components/chatbot/utils/markdownRenderer.tsx
```diff
@@ -14,14 +14,38 @@
...
+
+// Lazy-loaded so mermaid (~1.5 MB) is only downloaded when a message
+// actually contains a ```mermaid code block.
+const MermaidDiagram = lazy(() => import('../components/MermaidDiagram'));
+
+// Lazy-loaded so react-syntax-highlighter + prism language files are only
+// downloaded when the first fenced code block is rendered.
+const SyntaxHighlightedCode = lazy(() => import('../components/SyntaxHighlightedCode'));
+
+const CodeBlockFallback: React.FC<{ code: string; isDarkMode: boolean }> = ({ code, isDarkMode }) => (
```
bedanley, line 31: Review: Test markdown rendering with all plugins

bedanley (dismissed):

lambda/handlers/api_tokens/domain_objects.py
```diff
@@ -14,13 +14,13 @@
...
 
+from lisa.utilities.time import now_seconds
 from pydantic import BaseModel, Field, field_validator
-from utilities.time import now_seconds
 
 
 def default_expiration() -> int:
     """Calculate default token expiration (90 days from now)"""
-    return now_seconds() + int(timedelta(days=90).total_seconds())
+    return int(now_seconds()) + int(timedelta(days=90).total_seconds())
```
jmharold, line 23: [now_seconds](https://github.com/awslabs/LISA/blob/2d88570dc425a61ec83ddd680dceae7b14aaf3dd/lambda/utilities/time.py#L25) returns an int

lambda/handlers/repository/lambda_functions.py
```diff
@@ -205,15 +205,19 @@ def similarity_search(event: dict, context: dict) -> dict[str, Any]:
...
+    include_score = query_string_params.get("score", "false").lower() == "true"
+    repository_id = path_params.get("repositoryId")
+    collection_id = query_string_params.get("collectionId")
 
-    repository = get_repository(event, repository_id=repository_id)
+    if not isinstance(repository_id, str) or not repository_id:
+        raise ValidationError("repositoryId is required")
+    repo_id: str = repository_id
+
+    repository = get_repository(event, repository_id=repo_id)
```
jmharold: `repository_id` can be passed directly into `get_repository` without assignment to `repo_id`. The type will have been narrowed down to `str` after the `isinstance` check or and exception will have been raised

drduhe: Ack - fixing.

lambda/handlers/api_tokens/domain_objects.py
drduhe: Fixing - good catch.

drduhe: Actually - this was failing before this change on my linting with:
```
- hook id: mypy
- duration: 1.19s

lambda/handlers/api_tokens/domain_objects.py:23: error: Returning Any from function declared to return "int"  [no-any-return]
Found 1 error in 1 file (checked 237 source files)
```

estohlmann (approved):

PR #1066, opened 2026-06-05
feat/opensearch hybrid

gingerknight:
*Issue #, if available:*

### Summary
- Enables hybrid search for OpenSearch RAG repositories. Users selecting "Hybrid" mode get results from both semantic embedding and keyword matching via OpenSearch's inline `search_pipeline` body 
- Requires OpenSearch 2.13+.

#### Key behaviors:
- OpenSearch repos advertise `supportsHybridSearch: true`
- `hybrid_retrieve()` sends inline `normalization-processor` pipeline with configurable vector/lexical weights
- Per-citation similarity scores plumbed end-to-end (visible in message metadata)
- Chat UI: "RAG Settings" card with linked Slider+Input weight controls, preset buttons, capability gating
- Weight params (`vectorWeight`/`lexicalWeight`) 
- Admin toggle controls global hybrid availability

### How it works

When a user selects "Hybrid" search mode, the request flows through three layers:

**The backend** adds a new `hybrid_retrieve()` method to `OpenSearchRepositoryService`. Instead of the existing vector-only kNN query, it constructs a compound query combining BM25 (keyword) and kNN (semantic) legs, with a `normalization-processor` search pipeline sent inline in the request body. The weights controlling how much each leg contributes (e.g., 0.7 vector / 0.3 lexical) arrive as query string parameters, validated by a `HybridWeights` Pydantic model in the Lambda handler before reaching the service. A new `RetrieveResult` dataclass standardizes the return shape across all repository backends, carrying both the documents and metadata like `actual_mode_used` and per-document similarity scores.

**The chat UI** introduces a "RAG Settings" card inside the session configuration modal. When the selected repository supports hybrid and the admin toggle is on, users see a search mode dropdown. Choosing "Hybrid" reveals weight controls — paired sliders and number inputs that stay linked (adjusting one automatically sets the other so they sum to 1). Three preset buttons (Balanced, Semantic-heavy, Lexical-heavy) offer quick starting points. The weights are stored per-session and sent to the API on each RAG query.

**The Python SDK** (`lisapy`) exposes `vector_weight` and `lexical_weight` as optional parameters on `similarity_search()`, passing them as query string params to the REST API. SDK consumers who don't specify weights get the server-side defaults (0.7/0.3).
  
### Feature Boundaries

- Search-mode dropdown defaults to "Vector" when `hybridSearch` admin toggle is off
- `deriveRagSearchMode()` returns `'vector'` when admin flag disabled or repo doesn't support hybrid
- No silent fallback for OpenSearch hybrid. Requires 2.13+; errors propagate to the client
- No persistent infrastructure changes (no OpenSearch pipelines created, no new DDB tables)
#### Disabled Rag Controls based on Repo/Collection

Screenshot 2026-06-05 at 14 38 12

#### Enabled Rag Controls for OpenSearch and Admin Toggle enabled
Screenshot 2026-06-05 at 14 39 10

#### Related
- **Phase 1 (Bedrock KB):** PR #1044

### DIFF
--- a/lambda/shared/python/lisa/rag/services/opensearch_repository_service.py
+++ b/lambda/shared/python/lisa/rag/services/opensearch_repository_service.py
@@ -52,6 +52,97 @@ class OpenSearchRepositoryService(VectorStoreRepositoryService):
         """
         return True
 
+    def hybrid_retrieve(
+        self,
+        query: str,
+        collection_id: str,
+        top_k: int,
+        model_name: str,
+        include_score: bool = False,
+        bedrock_agent_client: Any = None,
+    ) -> tuple[list[dict[str, Any]], dict[str, Any]]:
+        """Retrieve documents using hybrid (BM25 + kNN) search via inline search pipeline.
+
+        Sends the search pipeline definition inline in the request body — no persistent
+        pipeline, no admin state, one round-trip. Requires OpenSearch 2.13+; older
+        clusters trip the narrow-exception fallback to vector search (slice 2.2.4).
+
+        Args:
+            query: Search query text
+            collection_id: Index/collection to search
+            top_k: Number of results to return
+            model_name: Embedding model name (used for query vector)
+            include_score: When True, copies hit['_score'] (already 0-1 from min_max
+                normalization) to metadata['similarity_score']
+            bedrock_agent_client: Unused for OpenSearch (kept for base-class signature parity)
+
+        Returns:
+            Tuple of (docs, retrieval_metadata) where retrieval_metadata reports
+            ``actual_mode_used='hybrid'`` and ``hybrid_supported=True`` on success.
+        """
+        embeddings = RagEmbeddings(model_name=model_name)
+        vector_store = self._get_vector_store_client(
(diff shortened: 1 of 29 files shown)


bedanley (dismissed):

lambda/shared/python/lisa/rag/services/opensearch_repository_service.py
```diff
@@ -43,6 +44,125 @@ class OpenSearchRepositoryService(VectorStoreRepositoryService):
     Only implements OpenSearch-specific index management.
     """
 
+    def supports_hybrid_search(self) -> bool:
+        """OpenSearch >= 2.13 supports hybrid search via inline search pipelines.
+
+        If the cluster is older than 2.13, hybrid_retrieve() will propagate the
+        resulting RequestError — no silent fallback.
+        """
+        return True
```
estohlmann, line 53: Seems strange to have a function that just returns `True`, especially when we call out cases that this would fail

estohlmann (approved):

